PR #24, opened 2024-04-02
Added landing page from dmsp_aws_prototype to app and dockerized the app

jupiter007:
I wanted to see if I could incorporate the landing page from dmsp_aws_prototype to the app, and use Apollo GraphQL client to make calls to the Apollo Server.

Additionally, I dockerized the app for dev and prod.


Dockerfile.dev
```diff
@@ -0,0 +1,26 @@
+# Dockerfile
+# preferred node version chosen here (LTS = 20.9.0 as of 10/24/2023)
+FROM node:20.9.0-alpine
```
briri, line 3: When we do builds in the AWS environment for this app, it's faster to use the AWS public ECR to fetch base images. Its also guaranteed (or vetted I guess) to work in their ECS container infrastructure

Can you switch this to: `FROM public.ecr.aws/docker/library/node:lts-alpine3.19`

Or if you prefer a different version you can use the select box at the top of this page to find one you want: https://gallery.ecr.aws/docker/library/node

briri: That image is accessible from your local machine, so you can use it for both local and AWS builds

Dockerfile.prod
```diff
@@ -0,0 +1,29 @@
+# Dockerfile
+# preferred node version chosen here (LTS = 20.9.0 as of 10/24/2023)
+FROM node:20.9.0-alpine
```
briri, line 3: See note on the dev version

README.md
```diff
@@ -2,35 +2,45 @@ This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next
...
-First, run the development server:
+### Installation
 
+#### Prerequisites
+- Docker
+- Node.js
+
+#### Development
+1. Clone the repository
 ```bash
```
briri, line 13: This is great. Thanks for documenting all of this

api.ts
```diff
@@ -0,0 +1,53 @@
+import { inDevMode } from "./utils";
+
+export class DmpApi {
+  getUrl() {
+    let apiHost = inDevMode() ? 'api.dmphub.uc3dev.cdlib.net' : `api.${window.location.hostname}`;
```
briri, line 5: This is good for now, but we should switch this to use GraphQL in the future. My intention is to let the old DMPTool use those API endpoints and for the new system to use GraphQL. So we will transition away from supporting both.

app/dmps/[...slug]/dmps.scss
Comment (briri) [inline app/dmps/[...slug]/dmps.scss:1]: I didn't even realize you could construct a directory name like this!

app/dmps/[...slug]/page.tsx
```diff
@@ -0,0 +1,330 @@
...
+    datasets: String[],
+    related_identifiers: RelatedIdentifier[],
+    versions: String[],
+}
+
+function getUrl() {
+    const headersList = headers();
+    const header_url = headersList.get('x-url') || "";
+    const url = new URL(header_url);
+    let apiHost = (url.hostname === 'localhost') ? 'api.dmphub.uc3dev.cdlib.net' : `api.${url.hostname}`;
```
Comment (briri) [inline app/dmps/[...slug]/page.tsx:64]: Isn't this duplicating the code in api.ts?

components/contributors.tsx
```diff
@@ -0,0 +1,124 @@
+import { Link } from '@/components/link/link';
+
+const ROLE_PREFIX_REGEX = /^https?:\/\/credit\.niso\.org\/contributor-roles\//i;
```
briri, line 3: Once we build out that ContributorRole table I will update the GraphQL schema to fetch the label and URL so that we can get rid of some of this logic :)

briri (changes requested):
Looks great @jupiter007!
I requested on small change to the image you're pulling for your Dockerfiles.

What is the difference between `.ts` and `.tsx`? Is it just that `.tsx` includes HTML?

jupiter007:
> Looks great @jupiter007! I requested on small change to the image you're pulling for your Dockerfiles.
> 
> What is the difference between `.ts` and `.tsx`? Is it just that `.tsx` includes HTML?

Thank you for looking at my PR @briri. The ".tsx" is equivalent to using ".jsx", which is used to indicate that we are using JSX in the component. ".ts" would be used like we would use ".js", which would be just pure javascript, but with type annotations.

Dockerfile.dev
jupiter007: @briri how can I tell what version of node that image contains?

jupiter007: oh nevermind. I think it's still the latest Long Term Support release version

app/dmps/[...slug]/page.tsx
Comment (jupiter007) [reply app/dmps/[...slug]/page.tsx:64]: Yea, it's a bit confusing. I was trying to see if I could generate the page on the server side, and the server side does not have access to the "window" browser object, so I needed to grab the same info about the host from the header.

components/contributors.tsx
jupiter007: Sounds good. I'll work on creating the Contributor Role Management page after I merge this.

jupiter007:
Hi @briri, I made the updates you requested. Please take another look when you get the chance and let me know if it looks ok. Thanks.

app/dmps/[...slug]/page.tsx
Comment (briri) [reply app/dmps/[...slug]/page.tsx:64]: ah ok. that makes sense. I had forgotten about the server side/client side division

briri (approved):
Thanks for updating the Dockerfiles

Issue #28, opened 2024-04-24
Frontend Infrastructure - File structure and shared methods for GraphQL requests


jupiter007:
For the DMPTool Rebuild frontend development, we are planning to use the latest version of NextJS 14 with App Router(i.e., React Server Components (RSC).

There are some nuances with the caching when using Apollo Client with RSC Server components in this version. Server Components must be manually refreshed when data changes. So if the same data is used in Client Components and Server Components, both environments might get out of sync. Apollo created experimental library that allows us to use Apollo Client in these RSC components: `@apollo/experimental-nextjs-app-support`. You can read more about it in these Apollo GraphQL blogs:

- https://www.apollographql.com/blog/how-to-use-apollo-client-with-next-js-13
- https://www.apollographql.com/blog/using-apollo-client-with-next-js-13-releasing-an-official-library-to-support-the-app-router

Apollo suggests that developers should choose whether to render Apollo Client data objects in RSC or in Client Components. When data is fetched from a GraphQL server, it is normalized into entities and stored in the cache. Subsequent queries can then retrieve data from the cache using the same entity structure, rather than refetching the data from the server. Therefore, when deciding whether to render each entity in RSC or in Client Components, it is important to consider how the data will be accessed and updated in the cache.

We need to install both `@apollo/client` and `@apollo/experimental-nextjs-app-support` and then choose between using Server Components or Client Components to make our GraphQL requests.

If we use GraphQL requests in Client Components, we can create one, shared client instance, `lib/apollo-client.ts`, and reference that client in our components to call on query or mutation functions saved at `lib/graphql/client/`.  That should work the same as it always has. If we need to use Apollo hooks in our Client Components, we might have to use the `ApolloWrapper` that is suggested in the above blog.

If we choose to make GraphQL requests in our Server Components, we can follow the instructions outlined in the above blogs, and create a `lib/client.js` file and use `getClient()` to make the requests. In order to refresh data on the server side, we have to use `revalidate` in our `fetchOptions`.

PR #36, opened 2024-04-30
Feature/graphql file structure

jupiter007:
I did some tests and made some updates per ticket [#28](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/28).

- Added some preliminary folder structure for graphql files and placed them under `lib/graphql`. This includes the client instance for graphql requests in Server Components, `client.ts`, client instance for Client components, `apollo-client.ts`, the `apollo-wrapper.ts` file used for the Apollo Provider to allow for GraphQL hookes in Client components, and mutations, queries, and methods to achieve those.
- I updated the main layout.tsx wrapper to include a placeholder Header and Footer
- I updated the root page.tsx to include placeholder data of collaborator roles which is done on the server side.
- I added some new packages, most importantly `@apollo/experimental-nextjs-app-support`
- I added a CHANGELOG.md file. I would like to keep records of changes there
- I removed references to tailwind


andrewebdev (commented):
Not much I can say here. All looks okay to me.
Something I did notive, which is a bit nit-picky I confess, is that in some places the indentation looks like 2 spaces, where in others it's four.

We should probably use "editorconfig" so that our editors match with the preferred coding style. (Though I admit this is low priority really, just a suggestion)

lib/graphql/server/queries/collaborator.ts
briri, line 1: This will depend largely on how you were hoping to organize code long term. Should this file be called `contributorRole.ts` instead to more accurately reflect what it is dealing with or were you intending to include the normal collaborator schema (email, name, etc.) in this file alongside these roles?

briri (approved):
This looks great @jupiter007. I have one minor suggestion about the name of a file.

briri:
> Not much I can say here. All looks okay to me. Something I did notive, which is a bit nit-picky I confess, is that in some places the indentation looks like 2 spaces, where in others it's four.
> 
> We should probably use "editorconfig" so that our editors match with the preferred coding style. (Though I admit this is low priority really, just a suggestion)

➕ for using editorconfig. It will be helpful

jupiter007:
> Not much I can say here. All looks okay to me. Something I did notive, which is a bit nit-picky I confess, is that in some places the indentation looks like 2 spaces, where in others it's four.
> 
> We should probably use "editorconfig" so that our editors match with the preferred coding style. (Though I admit this is low priority really, just a suggestion)

Thanks for bringing that up @andrewebdev. I totally agree. I think it's a good idea to use .editorconfig to set some standard configs. I will create a ticket for that. I have some extensions in VSCode that is probably contributing to the inconsistencies.

lib/graphql/server/queries/collaborator.ts
jupiter007: @briri - The file name would probably be better as contributorRole.ts, but I was including here more as a sample of what the code could look like and didn't think it would be a permanent file. In the root page.tsx I made a note that this is just a test to show how a GraphQL request for an RSC component might look like.

jupiter007:
I'm going to merge this PR since I believe I addressed the comments.

Issue #20, opened 2024-04-01
Contributor Role page


briri:
Create a contributor role management page (template for any lookup tables)
- Only super admin can access
- Display existing as a list with fields (label, URL, active checkbox)
- Ability to delete, and add new
- Will incorporate Apollo Client GraphQL to make queries and mutations

The initial entries to use will be from the [Credit NISO taxonomy](https://jats4r.niso.org/credit-taxonomy/). 
```
[
  { "id": 1, "label": "Data Manager", "url": "https://credit.niso.org/contributor-roles/data-curation/" },
  { "id": 2, "label": "Principal Investigator", "url": "https://credit.niso.org/contributor-roles/investigation/" },
  { "id": 3, "label": "Project Administrator", "url": "https://credit.niso.org/contributor-roles/project-administration/" }
]
```

jupiter007:
The Contributor Role page was created and the GraphQL calls to the development branch of the backend code was added and tested.

PR #38, opened 2024-05-06
Added some error handling ideas

jupiter007:
This research was done as part of issue [#20].
(https://github.com/CDLUC3/dmsp_backend_prototype/issues/20)

I made some updates as examples of how we can handle errors:

- For graphQL requests:
-- Updated `lib/graphql/client.ts` with adding an `errorLink` to the client instance, and intercepting specific error types or codes to handle them differently if we want.
-- Updated `lib/graphql/server/queries/dmpDataQueries.ts` to 1) Map the error code to a user friendly message using a hash like `utils/userFriendlyErrorMessage`, log the error to the new `utils/logger.ts`, and throw an error that will be caught by the "try-catch" in the client- or server-side component. `logger.ts` can be used directly in client- and server-components as well.
-- Updated `dmps/[...slug]/page.tsx` to throw the returned error in the 'try-catch'
-- Installed `pino` and `@elastic/ecs-pino-format` to output formatted logs
- Added `app/not-found.tsx` to handle 404 errors 
- Added `app/error.tsx` to handle server-side errors. Client components will most likely display returned errors in the page using useState().

To test:

1. Start up app
2. navigate to `http://localhost:3001/dmps/10.48321/D1SP4H`
3. Look at the stack trace, and you will see the log: `{"log.level":"error","@timestamp":"2024-05-06T23:13:42.224Z","process.pid":4920,"host.hostname":"CDL-JSHIN-9M","ecs.version":"8.10.0","loggerPath":"getDMSP","message":"Error fetching item from DynamoDB"}`


lib/graphql/client.ts
```diff
@@ -1,14 +1,51 @@
...
+                    console.error(`[GraphQL Error]: ${message}`)
+                    break;
+            }
+        })
+    }
+
+    if (networkError) {
+        const customNetworkError = networkError as CustomError;
+
+        customNetworkError.customInfo = { customMessage: 'There was a problem ' };
```
jupiter007, line 41: This is just displaying the possibility of manipulating data or handling a specific error code differently using `errorLink`

lib/graphql/server/queries/dmpDataQueries.ts
```diff
@@ -127,6 +129,18 @@ export async function getData() {
     };
     return formData;
   } catch (error: any) {
-    console.log(`Something went wrong: ${error.message}`)
+    const err = JSON.parse(JSON.stringify(error));
```
jupiter007, line 132: Here I am showing that we could extract error codes in the graphql request methods when there is an error, and map to a user-friendly message, log the error, and then throw the error. The error needs to be thrown here in order for it to be caught in the client-side or server-side component calling these methods.

utils/userFriendlyErrorMessages.ts
```diff
@@ -0,0 +1,12 @@
+export interface ErrorMessages {
```
jupiter007, line 1: This is just an example of how we could map specific error codes to user-friendly messages.

jupiter007:
Will go ahead and merge my changes, since I went over the changes with the team.

PR #44, opened 2024-05-14
added the buildspec.yaml file for the CI/CD pipeline

jupiter007:
I added a buildspec.yaml file to be used in the AWS Code Pipeline.

### DIFF
--- a/buildspec.yaml
+++ b/buildspec.yaml
@@ -1,0 +1,50 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      docker: 20
+
+  pre_build:
+    commands:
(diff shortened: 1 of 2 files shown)


buildspec.yaml
```diff
@@ -0,0 +1,50 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      docker: 20
```
jupiter007, line 6: I was reading that for docker containers, you want to specify the docker version instead of the node version. Not sure if that's correct or not.

This file is essentially the same as the one in dmsp_backend_prototype, except for the node references.

briri: Ah ok cool. I didn't realize that.

briri (approved):

PR #46, opened 2024-05-22
added healthcheck page for the application load balancer

briri:
The AWS load balancer requires a health check. The current homepage returns a 500 at the moment, so the ALB will not direct traffic to nextJS.

I have added a simple health check page here. I tried to override the default layout, so that it just returns something very simple like `<html><body>Ok</body></html>` but couldn't figure out how to do that.

Once this is merged to development, it should auto-deploy to ECS and the site should become available


jupiter007:
Thanks @briri. That makes sense to me to have a placeholder health check page for now.

jupiter007 (approved):

Issue #19, opened 2024-03-24
Implement Authentication in frontend 


jupiter007:
We should start building out how we want to handle authentication on the frontend.
1. Build placeholder login and signup pages 
2. After the backend returns tokens, save token in cookie securely using 'httponly'
3. Will need to add middleware to verify token to see if user is authenticated before allowing users to some protected pages

jupiter007:
I was able to authenticate with Cognito in the NextJS app, within client-side components.

I created a client-side "sign-up" page using the following packages: "aws-amplify" and "aws-ampfliy/auth".  

I created the following amplify.js config component, 
```
import {Amplify} from "aws-amplify";

const awsConfig = {
  Auth: {
    Cognito: {
      userPoolId: process.env.NEXT_PUBLIC_USER_POOL_ID,
      region:process.env.NEXT_PUBLIC_REGION,
      userPoolClientId: process.env.NEXT_PUBLIC_POOL_APP_CLIENT_ID,
      loginWith: {
        username: 'true'
      }
    }

  }
} 

export default awsConfig
```
and configured Amplify with it:

```
'use client';
import React, {useState, useEffect} from 'react';
import {Amplify} from "aws-amplify";
import {useRouter} from 'next/navigation';
import awsConfig from '@/amplify';
import {signUp,confirmSignUp} from "aws-amplify/auth";
import ConfirmUser from '@/components/confirmation';

Amplify.configure(awsConfig as any)


  const SignUp:React.FC = () => {
    const [email, setEmail] = useState("");
    const [password, setPassword] = useState("");
    const [newUser, setNewUser] = useState(false);
    const [verificationCode, setVerificationCode] = useState("");
    const router = useRouter();

    const handleSignUp = async(event: React.FormEvent) => {
      event.preventDefault();
      try {
        const {isSignUpComplete, userId, nextStep} = await signUp({
          username: email,
          password: password,
        })
        console.log("User registered:");
        if(userId) {
          setNewUser(true);
        }
      } catch (error) {
        console.error("Error registering user:", error)
      }
    };

    const handleConfirmSignUp = async (e: React.FormEvent) => {
      e.preventDefault();
  
      try {
        const response = await confirmSignUp({username:email, confirmationCode: verificationCode});
        console.log(response);
        router.push('/login');
      } catch (err) {
        console.error(err);
      }
    }
    return (
      <main className="bg-gray-200 h-screen flex items-center justify-center">
[...]

PR #56, opened 2024-06-12
Added placeholder /login and /signup page to handle access token and adding /api endpoint to create cookie

jupiter007:
Ticket: #19 

1. Added /login and /signup placeholder pages to test that I could get access token from the backend and save the access token in a cookie
2. Added initial middleware to verify access token to check if user is authenticated
3. Added unit tests and jest configurations for the tests


app/signup/__tests__/page.spec.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import SignUpPage from '../page';
+
+jest.mock('next/navigation', () => ({
+    useRouter: jest.fn()
+}));
+
+//Need to import this useRouter after the jest.mock is in place
```
briri, line 9: Interesting, I think this may have been what was tripping me up in my tests.
So this is essentially mocking the entire `next/navigation` module and then you are importing that "mocked" version of the `userRouter` function?

package.json
```diff
@@ -15,11 +16,15 @@
     "@fortawesome/fontawesome-svg-core": "^6.5.2",
     "@fortawesome/free-solid-svg-icons": "^6.5.2",
     "@fortawesome/react-fontawesome": "^0.2.0",
+    "@types/cookie": "^0.6.0",
```
briri, line 19: You can probably move these `@types/*` to your devDependencies

briri (approved):
This looks great @jupiter007. Just 2 minor changes, feel free to merge afterward

app/api/setCookie/route.spec.ts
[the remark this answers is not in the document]
jupiter007: Good point. Yes we should be consistent. I'll move it.

app/signup/__tests__/page.spec.tsx
jupiter007: Yes that is my understanding. Mocking "useRouter" took me forever to figure out. It was a big headache, but I finally got it to work by importing after the jest.mock.

package.json
jupiter007: Thanks Brian. I will make that change.

Issue #55, opened 2024-06-11
Add access token to GraphQL calls as "Authorization: Bearer <access token>"


jupiter007:
Update Apollo Client creation code to include a 'authLink' that will always add the JWT token to GRAPHQL calls?

mariapraetzellis:
@jupiter007 is working on updates to the middleware to handle the authentication & will add unit tests.

PR #60, opened 2024-06-24
Added access tokens to GraphQL request headers, updated middleware to handle authorization

jupiter007:
Issue: #55 

- Added JWT token to headers for graphql requests on both the client- and server-side components. Updated `lib/graphql/client.ts` and `lib/graphql/apollo-wrapper.tsx`
- Added an example client-side component called 'client-graphql-test' to show how apollo client hooks can work on client side using the 'apollo-wrapper'.
- Updated middleware.ts for authorization and redirection and updated middleware spec
- Added a 'get-token' endpoint to get the token from the cookie and added unit test
- Updated auth.ts to use 'jose' instead of 'jsonwebtoken' for token verification due to errors related to the edge runtime not support Node.js 'crypto' module.
- Added utils:
--  `utils/authLink.ts` - Creates 'authLink' for the creation of the client instances
--  `utils/cookiesUtil.ts` - Contains a `deleteCookie` method
--  `utils/getAuthTokenServer.ts` - Gets JWT token from cookie

**To Test authentication and authorization**:
- Run development branch of `dmsp_backend_prototype`. (Make sure the JWT secret is the same for backend as frontend)
- Run this frontend branch and go to `/signup`. Signup with a new user. 
- Go to `/login` and sign in with new user. Notice the new '`dmspt`' httpOnly cookie that was added
- Navigate to `/dmps/10.48321/D1SP4H`. There are currently errors getting data for the page, but you should be able to acces that page.
- Delete the '`dmspt`' cookie and try navigating to `/dmps/10.48321/D1SP4H`. You should be redirected to /login page

**To Test Token in Header for Client Side**:
- Navigate to `/client-graphql-test` and look at your Network tab and filter for graphql. You should see the token in the Authorization header


lib/server/auth.ts
```diff
@@ -1,23 +1,31 @@
-import jwt, { Secret } from 'jsonwebtoken';
+import { jwtVerify, JWTPayload } from 'jose';
 
-export function getJwtSecretKey(): Secret {
-    // Secret is hardcoded for now
-    // TODO: Need to figure out where the secret will be stored
-    const secret = Buffer.from('Zn8Q5tyZ/G1MHltc4F/gTkVJMlrbKiZt', 'base64');
+/**
```
jupiter007, line 3: Changed to using 'jose' package to verify the token. This is because the "Edge runtime environment does not include the 'crypto' Node.js core module, and jsonwebtoken relies on this module for cryptographic operations.

The 'jose' package is designed to work without this, using standard APIs like Web Cryptography API which is supported in browsers.

__tests__/middleware.spec.ts
```diff
@@ -5,101 +5,94 @@
...
-    it('should redirect to login if no token is found for protected routes', () => {
-        const request = mockNextRequest('/dmps/123');
-        middleware(request);
+    it('should redirect to login if no token is found for protected path', async () => {
+        (getAuthTokenServer as jest.Mock).mockResolvedValue(null);
+        const request = new NextRequest(new Request('http://localhost:3000/dmps/123'));
+        const response = await middleware(request);
 
-        expect(redirectSpy).toHaveBeenCalledWith(`${process.env.NEXT_PUBLIC_BASE_URL}/login`);
+        expect(redirectSpy).toHaveBeenCalledWith('http://localhost:3000/login');
```
andrewebdev, line 41: Just wondering if there is a reason to hard-code the localhost url here? Keeping it as an environment variable might be preferable, especially since every dev might have their environment set up differently.

__tests__/middleware.spec.ts
```diff
@@ -5,101 +5,94 @@
...
+        (getAuthTokenServer as jest.Mock).mockResolvedValue('invalid_token');
+        const request = new NextRequest(new Request('http://localhost:3000/dmps/123'));
         (verifyJwtToken as jest.Mock).mockImplementation(() => false);
-        middleware(request);
+        await middleware(request);
 
-        expect(request.cookies.delete).toHaveBeenCalledWith('dmspt');
-        expect(redirectSpy).toHaveBeenCalledWith(`${process.env.NEXT_PUBLIC_BASE_URL}/login`);
+        expect(deleteCookie).toHaveBeenCalled();
+        expect(redirectSpy).toHaveBeenCalledWith('http://localhost:3000/login');
```
andrewebdev, line 51: Same as previous comment on hard-coded url. There are more in this document, but I'm not going to flag each individual one :)

__tests__/middleware.spec.ts
andrewebdev: Ah, reading further into the code, looks like all of this is inside the context of a test and mocking the request/response. If this is the case then it's understandable to use a fixed url.

app/api/get-token/__tests__/route.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+        const data = await response.json();
+
+        expect(data).toEqual({ token: null })
+    })
+
+    it('should handle internal server error', async () => {
+        (getAuthTokenServer as jest.Mock).mockRejectedValue(new Error('Test error'));
+
+        const response = await GET();
+        expect(response.status).toEqual(405);
```
andrewebdev, line 40: `405: Method not Allowed` doesn't seem to be the correct error code for this case. The test doesn't seem to check if the correct method (GET, POST etc.) is used.

Further down, the error message also reads as a `Internal Server Error` which should be a `500` code. So it feels like things aren't matching up here.

app/api/get-token/route.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const GET = async () => {
+    try {
+        const token = await getAuthTokenServer();
+
+        return NextResponse.json({ token }, { status: 200 })
+
+    } catch (err) {
+        console.error('Error fetching token:', err);
+        return NextResponse.json({ error: 'Internal Server Error' }, { status: 405 });
```
andrewebdev, line 12: From the use case here it seems like `500` should be the correct code to use. 

What would be the cases where a token cannot be found? We probably need to handle each of those cases with different error responses.

If the token cannot be found cause a fault on our side, then `500` would be the correct code. But if it cannot be found cause the user isn't signed in, then `401` is probably the correct code.

But most likely the status `405` here is not the right one, since the method is technically allowed, so long as you have a token.

lib/server/auth.ts
```diff
@@ -1,23 +1,31 @@
...
+ * Get the JWT Secret
+ * @returns 
+ */
+export function getJwtSecretKey(): Uint8Array {
+    const secret = process.env.JWT_SECRET
```
andrewebdev, line 8: _Just a side note:_ We should probably set up a mechanism somewhere to easily rotate  secret keys if needed. It's good practice to rotate them every so often to improve site security.

@briri , we should probably create an issue for this so we are reminded of it at a later stage?

briri: My thought was to store this and probably other variables we have yet to identify in AWS Secrets manager which you can configure to rotate on a schedule. 

@jupiter007 can we move this `process.env.JWT_SECRET` to its own function for now? That will make it easier to update when we do swap in Secrets Manager.

app/api/get-token/route.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+};
+
+export const POST = () => {
```
briri, line 16: these are great. I like that we explicitly block them

__tests__/middleware.spec.ts
jupiter007: Yes, @andrewebdev as you mentioned above, I mocked the request and passed in the URL so that I could check the redirect url.

briri (approved):
this all looks good to me

app/api/get-token/__tests__/route.spec.ts
jupiter007: Thanks @andrewebdev. The 500 error code totally makes sense to me since it's on the server side. I think I was copying and pasting and forgot to change that one.

app/api/get-token/route.ts
jupiter007: Thanks @andrewebdev. I will change it to statusCode of 500.

lib/server/auth.ts
jupiter007: Thanks @andrewebdev and @briri. Using AWS Secrets Manager is a great idea. I'll add a placeholder function for now.

jupiter007: I'll also create a ticket to use Secrets Manager in the NextJS app.

jupiter007: I created this ticket to implement Secrets Manager on the frontend: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/62

@briri should I create another ticket for setting up AWS Secrets Manager?

jupiter007:
@andrewebdev and @briri thanks so much for your code reviews. I really value your input. I've made the changes you requested so far. Please let me know if there is anything else you recommend that I update or add. I'm sure we'll discover that we need more changes as we progress with the project as well.

jupiter007:
Hi @briri and @andrewebdev, I consolidated my es lint configs into .eslintrc.json and fixed some linting issues that weren't caught with "npm lint". For now, I added "npx eslint ." to the "lint" script until I can figure out how to capture all directories.

I also updated some package versions.

Issue #66, opened 2024-07-06
Add loading and velocity control on signup and signin forms


jupiter007:
We need to add velocity control on form submits to prevent `repeated submit attacks`, and also add loading states to help with that and for visual feedback.

PR #72, opened 2024-07-15
Added logging state and velocity controls to the login and signup forms

jupiter007:
## Description

We need to add some controls on the frontend to stop users from repeated submit attacks. I implemented a loading state to disable the "submit" button while the POST request is being made. That way a user is prevented from repeatedly pushing the button to start new requests while the request is in progress.

I also added an "attempts" state. After 5 attempts, the user gets an error message telling them that there were too many attempts, and that they need to wait before tying again.

The messaging, and possibly some of the functionality, will be updated when the designs for these pages come in. However, I just wanted to put some minimal defenses in place to remind us to address these with the finalized pages.

Fixes # ([#66](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/66)

## Type of change

## How Has This Been Tested?
The associated unit tests were updated to test some basics of the new functionality.

1) Start the app
2) Navigate to either the /login or /signup page
3) Make more than 5 attempts to submit bad data in the form
4) On the sixth try, you should get the "too many attempts" message which should display the amount of minutes the user has to wait to try again.

## Checklist:
- [ ] My code follows the style guidelines of this project (Since the signup and login form pages are just placeholders, and no final style guides are in place, this is not the case.
- [X] I have completed manual or automated accessibility testing for my changes (I did not carry out detailed accessibility tests, since these pages are just placeholders until we get a final design in place)

### DIFF
--- a/app/login/page.tsx
+++ b/app/login/page.tsx
@@ -17,6 +17,8 @@ const LoginPage: React.FC = () => {
         password: "",
     });
     const [errors, setErrors] = useState<string[]>([]);
+    const [loading, setLoading] = useState(false);
+    const [attempts, setAttempts] = useState(0);
     const router = useRouter();
     const errorRef = useRef<HTMLDivElement>(null);
 
@@ -70,6 +72,14 @@ const LoginPage: React.FC = () => {
 
     const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault();
+        setLoading(true);
+        setErrors([]); // Clear previous errors
+
+        if (attempts >= 5) {
+            setErrors(['Too many attempts. Please wait before trying again.']);
+            return;
+        }
+
         /* eslint-disable @typescript-eslint/no-explicit-any */
         try {
             const response = await fetch(`${process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT}/signin`, {
@@ -87,6 +97,9 @@ const LoginPage: React.FC = () => {
                 error: err,
                 url: { path: '/signin' }
             });
+        } finally {
+            setLoading(false);
+            setAttempts(attempts + 1);
         }
 
     };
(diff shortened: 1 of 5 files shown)


app/login/page.tsx
```diff
@@ -70,6 +72,14 @@ const LoginPage: React.FC = () => {
 
     const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault();
+        setLoading(true);
+        setErrors([]); // Clear previous errors
+
+        if (attempts >= 5) {
+            setErrors(['Too many attempts. Please wait before trying again.']);
```
briri, line 79: If we are setting it to 15 minutes, we should be specific in the message

app/signup/page.tsx
```diff
@@ -36,8 +38,14 @@ const SignUpPage: React.FC = () => {
 
     const handleSignUp = async (event: React.FormEvent) => {
         event.preventDefault();
+        setLoading(true);
         setErrors([]); // Clear previous errors
 
+        if (attempts >= 5) {
+            setErrors(['Too many attempts. Please wait before trying again.']);
```
briri, line 45: see comment above

app/login/page.tsx
```diff
@@ -130,7 +153,7 @@ const LoginPage: React.FC = () => {
                     onChange={handleInputChange}
                     required
                 />
-                <button type="submit">Login</button>
+                <button type="submit" disabled={(loading || attempts > 5) ? true : false}>{attempts}{loading ? 'Logging in ...' : 'Login'}</button>
```
briri, line 156: Looks like you're always showing the number of attempts here. Is that for debug?

briri (approved):

app/login/page.tsx
jupiter007: @briri do you think 15 minutes makes sense to you? It was just a random number.

app/login/page.tsx
jupiter007: Oh thanks for catching that Brian :) Yes, that was for debugging.

app/login/page.tsx
jupiter007: @briri I updated the code to use a more dynamic approach to lockout times, making it more difficult for bad actors to commit brute force attacks. I also included the remaining minutes in the message to the user.

briri: Yeah, I think 15 makes sense. Thanks for making it dynamic :)

Issue #65, opened 2024-07-06
Add logout button


jupiter007:
This ticket is blocked by: https://github.com/CDLUC3/dmsp_backend_prototype/issues/43

Once the above is complete, then add a logout button that calls that endpoint, and subsequently deletes the JWT token cookie.

bofstein:
Start with removing cookie rather than waiting on #43

jupiter007:
Switched back to "In Progress" since I want to add some handling of context for authentication.

PR #73, opened 2024-07-17
Feature/add logout

jupiter007:
## Description

Fixes # [#65](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/65)

1. Added a "logout" link and API endpoint to delete the "dmspt" cookie when the user logs out.
2. Added an AuthProvider to wrap around the content so that we can share the authentication state in the client components, and update when the user logs in or logs out.
3. Since I added an AuthProvider, I had to create a custom render function for the unit tests of client components that use the AuthProvider.

Eventually, we should also be calling the backend server so that the JWT token can be invalidated.

## Type of change

## How Has This Been Tested?

1. Start the app
4. Go to http://localhost:3000/login and log in
5. Notice the "dmspt" cookie after you log in
6. Notice that the link text in the header says "Log out" after you successfully log in
8. Click on the "Log out" link in the header
9. You should be redirected to the "/login" page and the cookie should've deleted
10. Notice that the link text in the header now says "Log in" after you successfully logged out.
11. Notice that the "dmspt" cookie has been deleted.

## Checklist:
- [] My code follows the style guidelines of this project (Not a final design. I just inserted a link into the header for testing purposes.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [1.0.0]
 
 ### Added
+- Added logout api and link [#65]
 - Added loading state and velocity controls to the login and signup forms [#66]
 - Added placeholders for /signup and /login pages and added a new api endpoint for setting the access token in a cookie [#19]
 - Added .editorconfig file to help maintain consistent coding styles [#37]
--- a/app/api/check-auth/__tests__/route.spec.ts
+++ b/app/api/check-auth/__tests__/route.spec.ts
@@ -1,0 +1,82 @@
+/**
+ * @jest-environment node
+ */
+import { NextResponse } from 'next/server';
+import { GET } from '../route';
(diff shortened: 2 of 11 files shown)


briri (approved):
This looks good to me Juliet

jupiter007:
@briri I made an update to use a context wrapper around the content, so that the client components can easily grab the authentication state, and update it when the user logs out or logs in.

briri (approved):
Thanks @jupiter007 I think this will be a good way to handle the auth state. the AuthContext also seems like a good spot to add an `isAdmin` state or something similar when we get around to permissions

jupiter007:
> Thanks @jupiter007 I think this will be a good way to handle the auth state. the AuthContext also seems like a good spot to add an `isAdmin` state or something similar when we get around to permissions

Yes. Also hoping to use middleware to control routing for different roles.

Issue #10, opened 2023-12-18
Build component for searching for Orgs


briri:
Build out a Typeahead component that allows a user to call OpenSearch to find Orgs that match the user's entry as they type.

Refer to the work already done in the DMPTool repo's `react-client`. This will need to be updated though to call OpenSearch.

We will need a broad one that searches across all orgs and one that is more refined to search by type (e.g. funder or non-funder)

mariapraetzellis:
This requires that the OpenSearch API is connected, which is tracked here: https://github.com/CDLUC3/dmsp_aws_prototype/issues/84

bofstein:
Built something for pilot project to transfer from
Requires the BE table to be able to build from, needs the endpoints
Dependent on this ticket: https://github.com/CDLUC3/dmsp_aws_prototype/issues/85

bofstein:
Also needs https://github.com/CDLUC3/dmsp_aws_prototype/issues/84 to get the newest org data

bofstein:
![image](https://github.com/user-attachments/assets/b0825e0f-505e-4eaa-8979-237352966179)

This version looks good to me if we fix the spacing around the first parentheses, rather than using a separate checkbox for Other, as that might be covered when typing.

If it's too long we can change it later to just "Organization not listed"

jupiter007:
Ok thanks @bofstein. I will remove the TypeAhead component that does not include "Other" then.

bofstein:
Now I will say I'm not sure if there's anywhere else on the site we might use a typeahead without a pinned Other option if it's a component we'd re-use.

jupiter007:
Ok, well maybe I'll keep the other one just in case, for now, and just change the "Other" language to the one you proposed above.

PR #87, opened 2024-08-12
Added two new TypeAhead versions to the styleguide

jupiter007:
## Description

We will need TypeAhead/Autosuggest components for pages like the "Register" page where users have to search for an Institution.

- I added two TypeAhead components to the app: the first one does not include an "Other" option and the second one has it, and expects a "setOtherField" method to be passed to it that can be set to true when a user selects the "Other" field. That way the calling component can know to show the additional "Other" field.

- I remove the files related to the pilot project dmp landing page because it was breaking the build, and was added primarily for testing purposes.
- I added the use of graphql-codegen to generate the types for the graphql queries.
- Updated README.md file

Fixes # [10](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/10)

## Type of change

## How Has This Been Tested?
The new TypeAheads were tested against the backend server GraphQL API for `affiliations` after they were added to the styleguide.

- Start the app along with the dmsp_backend_prototype `development` branch.
- Navigate to http://localhost:3000/styleguide and scroll down to the `Forms fields` section.
- Some tests to run
-- Enter "alaska" into input field. Should show results in dropdown
-- Select an option from the dropdown either using "Click" or "Enter" key. The input value to change to the selected one
-- Navigate through the dropdown using the up and down arrow keys
-- Enter a nonsensical value like "asdfadf". Try and tab out of the field. You should be restricted from doing so.

Also, unit tests for the components were added.

## Checklist:


app/client-graphql-test/page.tsx
```diff
@@ -13,6 +11,11 @@ query ContributorRoles{
     }
 }`;
 
+type ContributorRoles = {
```
briri, line 14: Could you setup codegen instead that auto-generates the types for you based on the GraphQL schema definition in the backend?
example `codegen.ts`
```
import type { CodegenConfig } from "@graphql-codegen/cli";

const config: CodegenConfig = {
  schema: "../dmsp_backend_prototype/src/schemas/*.ts",
  generates: {
    "./src/types.ts": {
      plugins: ["typescript", "typescript-resolvers"],
      config: {
        contextType: "../dmsp_backend_prototype/src/context#MyContext",
        enumsAsTypes: true,
      },
    },
  },
};

export default config;
```

components/Spinner/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    className: string;
+}
+function Spinner(props: SpinnerProps) {
+    return (
+        <>
+            {props?.isActive && (
+                <div
+                    id={(props.id ? props.id : "spinner")}
+                    tabIndex={-1}
+                    aria-hidden="true"
```
briri, line 16: this is just a question, but how does a screen reader let the user know that the site is loading/processing?

briri (approved):
This looks good to me @jupiter007 

I just had one suggestion about using codegen to import the types from the backend codebase.

A we going to allow for both Typeaheads or are we waiting for consensus on which one is preferred?

app/client-graphql-test/page.tsx
jupiter007: @briri Yes that's a good idea. I'll create a separate ticket for that.

components/Spinner/index.tsx
jupiter007: Good point. I should probably add some hidden text that is readable by the screen reader that says "loading".

app/client-graphql-test/page.tsx
jupiter007: @briri I created this ticket to add the use of graphql-codegen: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/88

Also, I added a "Loading..." message for screen readers in the TypeAheads.

jupiter007:
@briri I merged the graphql-codegen changes to this PR. Could you take a quick look at the last two commits when you get the chance? Thanks.

Issue #89, opened 2024-08-15
Create overall DMP Tool Header


bofstein:
Should look the same as header on the current page. Likely need to fully rebuild this since it was a Rails template before.

![Image](https://github.com/user-attachments/assets/0a1f738e-e19b-4184-89aa-17cfc9ed9947)

Note that the Admin button is only for Admins

The drop-downs are currently hard to click on because a tiny bit of mouse movement outside the box makes it go away, something to improve in the new version
![Image](https://github.com/user-attachments/assets/250452dc-3dae-4e88-9201-1c8decfcbc25)

May need design input on dropdown, e.g. from Admin menu

jupiter007:
Something to note is that this header and footer use "fontawesome" icon sets. Although I've copied the styles over in the "feature/add-header-and-footer" branch, I'll discuss with the developer team on Tuesday to see whether there's a point to putting the existing header and footer into the app, due to lots of style changes.

PR #90, opened 2024-08-16
Added existing header and footer to the app

jupiter007:
## Description

For now, I added the existing header and footer from production into the app so that we can get a sense of the look and experience as we start developing the new pages.

I realize that we will be making style changes. There are things like the icon library that we will be changing over from fontawesome to Google Material, but it might be easier to see what we might want to change by having the existing header and footer in place. Of course, we'll have to clean up the classes if we want to stick with this layout, and also we will need to add new tickets to improve the dropdown menus in the navigation.

Fixes # [89](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/89)

## Type of change
Moved header and footer from production into the app. We are assuming that most of the header and footer design will remain the same, but if not, at the very least it gives us a better sense of how the new page designs will work with the header and footer, and what we will want to update/change.

image


app/dmps/[...slug]/dmps.scss
```diff
@@ -1,464 +1,464 @@
```
Comment (briri) [inline app/dmps/[...slug]/dmps.scss:1]: Why not just delete this file? It looks like everything is now commented out.

briri (approved):
Looks good @jupiter007

app/dmps/[...slug]/dmps.scss
Comment (jupiter007) [reply app/dmps/[...slug]/dmps.scss:1]: @briri it was deleted in the other TypeAhead branch already, so I didn't bother doing it here

Issue #86, opened 2024-08-12
Build Changelog history page


bofstein:
Make page to view history of published changes to template based on table from https://github.com/CDLUC3/dmsp_backend_prototype/issues/14 (mock data to start)

![Image](https://github.com/user-attachments/assets/90b86bc5-62d7-4e08-b991-3b18d92879ca)

bofstein:
Backend should have some data to work with

PR #101, opened 2024-08-26
[Feature] - Built template history page

jupiter007:
## Description

Built the Template History page. 

- Created a Back Button component to share across different pages
- Made sure to set page header, and subheader info based on the most recent publication
- Added versions in descending order
- Added a new `_variables.scss` file that contains definitions for breakpoints and a new `_mixins.scss` file to be able to set styles for different devices. For now, I only added support for small devices and devices that are tablets and larger. We can update as needed to include more devices
- I understand that we don't have the final styles, but for now, I added styles to shared `_typography.scss` and `_page.scss` files for shared styles, which can be updated when final design and styleguide is complete.
- Added a PageWrapper component to wrap all pages so that we can add title tags to pages, and allow for scrolling to top of page
- Added shared dateUtils functions for formatting data

Fixes # ([86](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/86))

## Type of change

## How Has This Been Tested?
A unit test was created to test all the basic functionality of the new page.

## Checklist:


app/template/[templateId]/history/page.tsx
```diff
@@ -0,0 +1,105 @@
...
+        { variables: { templateId } }
+    );
+
+    // Handle loading state
+    if (loading) {
+        return <p>Loading publication history...</p>;
+    }
+
+    if (error) {
+        return <p>There was a problem.</p>
```
Comment (jupiter007) [inline app/template/[templateId]/history/page.tsx:26]: L25: I will add a shared error handler in a separate ticket.

app/template/[templateId]/history/page.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const params = useParams();
+    const templateId = Number(params.templateId);
+
+    const { data = {}, loading, error } = useTemplateVersionsQuery(
+        { variables: { templateId } }
+    );
+
+    // Handle loading state
+    if (loading) {
+        return <p>Loading publication history...</p>;
```
Comment (jupiter007) [inline app/template/[templateId]/history/page.tsx:22]: L22 - I'm planning to add a loading spinner as part of a separate ticket.

components/BackButton/index.tsx
```diff
@@ -0,0 +1,13 @@
+'use client';
+
+import { useRouter } from 'next/navigation';
+
+export default function BackButton() {
+    const router = useRouter();
+
+    return (
+        <button className="back-link-button" onClick={() => router.back()}>
+            &larr; Back
```
jupiter007, line 10: L10: Just used the arrow element until we have the Google Material symbols available.

app/template/[templateId]/history/page.tsx
```diff
@@ -0,0 +1,105 @@
...
+                                            <td>{publishDate}</td>
+                                        </tr>
+                                    );
+                                })
+                                : <tr><td colSpan={3}>No template history available.</td></tr>
+
+                        }
+
+                    </tbody>
+                </table>
```
Comment (andrewebdev) [inline app/template/[templateId]/history/page.tsx:99]: Should we maybe be using the react-aria table for this instead?
See: https://react-spectrum.adobe.com/react-aria/Table.html

styles/_mixins.scss
```diff
@@ -0,0 +1,12 @@
...
+    @if $breakpoint == sm {
+        @media (max-width: $breakpoint-sm) {
+            @content;
+        }
+    } @else if $breakpoint == lg {
+        @media (min-width: $breakpoint-lg) {
+            @content;
+        }
+    }
+}
```
andrewebdev, line 12: I saw your discussion in slack with @fraserclark on this.
I actually already created a bunch of responsive helpers for this in; `@/components/responsive`.

Regarding @fraserclark 's point on not wanting to use names like "tablet", I'm not very opinionated on this, but would be interested to hear the reasoning behind it. My only comment regarding this would be that I normally prefer names that I can recognize, and don't have to go look up what the "shortcode" for the size is.

styles/_typography.scss
```diff
@@ -57,3 +56,19 @@ code {
   color: var(--brand-tertiary);
   font-weight: bold;
 }
+
+
+@media only screen and (min-width: 600px) {
```
andrewebdev, line 61: For consistency we should probably use the media query mixins for this?

styles/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$breakpoint-sm: 600px; //Mobile (small devices)
+$breakpoint-lg: 601px; //Large devices like desktop
```
andrewebdev, line 2: Regarding my earlier comment on why I prefer semantically named sizes for breakpoints, the fact that a comment is needed to indicated what size this refers to is, for me at least, an indication that the names should be more semantic.

utils/dateUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+export const formatShortMonthDayYear = (isoString: Date) => {
+    if (isoString) {
+        const date = new Date(isoString);
+        const options: Intl.DateTimeFormatOptions = { year: 'numeric', month: 'short', day: 'numeric' };
+        return date.toLocaleDateString('en-US', options);
+    } else {
+        return 'No date';
+    }
+
+}
```
andrewebdev, line 36: Is there a reason we aren't using something like "momentjs"[1] for this? It makes working with date and time much easier and consistent.

[1] https://momentjs.com/docs/

andrewebdev (commented):
Other than the couple of comments I made, everything else looks good to me.

utils/dateUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        // Define arrays for month names
+        const months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+
+        // Extract date components
+        const hours = date.getUTCHours();
+        const minutes = date.getUTCMinutes();
+        const day = date.getUTCDate();
+        const month = date.getUTCMonth(); // Month is 0-based
+        const year = date.getUTCFullYear();
```
briri, line 14: Should we use locale time in the UI instead of UTC? I suspect that most users would prefer it

briri (approved):

utils/dateUtils.ts
briri: yes agree with this.

public/images/bg-arrow-Up.svg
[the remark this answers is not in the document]
jupiter007: @andrewebdev it's a background image for the mobile menu

utils/dateUtils.ts
jupiter007: @andrewebdev and @briri, do we need to internationalize dates in the app? If not, I think, for now, we can leave this dateUtils.ts in place, and discuss other options as we progress in the project.

I read up on Moment, and the docs for Moment are discouraging developers from using it because it is outdated and because of its size. (https://momentjs.com/docs/).  The site also mentions that there is a proposal to include "[Temporal](https://tc39.es/proposal-temporal/docs/index.html)" in the JavaScript library to handle dates and times.

utils/dateUtils.ts
jupiter007: Oops. Thanks for catching that @briri .

app/template/[templateId]/history/page.tsx
Comment (jupiter007) [reply app/template/[templateId]/history/page.tsx]: @andrewebdev - Yes thanks. I converted it to use React Aria Table, and also added it to the styleguide.

One weird thing is that the <Cell> component doesn't include "colspan". Sometimes these types of libraries have some weird quirks.

styles/_mixins.scss
jupiter007: @andrewebdev and @fraserclark, I went ahead and switched to using the breakpoint labels in responsive.scss. Honestly, I totally forgot you had added those breakpoints already Andre. Do you think we should move that file out of components and into the "/styles" directory? It seems to make more sense to me than keeping it in components.

styles/_typography.scss
jupiter007: Thanks @andrewebdev. Forgot to update that one with the mixins.

app/template/[templateId]/history/page.tsx
Comment (andrewebdev) [reply app/template/[templateId]/history/page.tsx]: We should also copy the styles from their components. @fraserclark and I have been doing that. We just copy the CSS from their examples, you just expand the css bits to get to it. We can later customize the styling from there, but it's a good starting point.

styles/_mixins.scss
andrewebdev: yeah you can move it to where it makes more sense. I don't recall my reasoning why I put it in that directory to begin with, but you are right it doesn't make sense there, and is probably why you didn't realize it's there already ;)

utils/dateUtils.ts
andrewebdev: Oh that's news to me! Thank you @jupiter007 ! I'll go read up on this and look at our options. I'm still of the opinion that we should avoid working with datetime ourselves and rather let a good library take care of it. I refer to that YT video by Tom Scott [1] :)

[1] https://www.youtube.com/watch?v=-5wpm-gesOY

public/images/bg-arrow-Up.svg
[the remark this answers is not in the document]
andrewebdev: We should probably use the material symbols for this as well?
I can probably extract my material symbols integration into a separate PR that we all can use it sooner, rather than wait for the remirror stuff to include them? Would that help?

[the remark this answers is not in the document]
jupiter007: @andrewebdev that makes sense to integrate material symbols soon, rather than later, since we're getting going on building pages now. Thanks.

styles/_mixins.scss
jupiter007: Thanks @andrewebdev. I moved it under "/styles"

app/template/[templateId]/history/page.tsx
Comment (jupiter007) [reply app/template/[templateId]/history/page.tsx]: Thanks @andrewebdev. I copied the styles over to `_elements.scss` and applied the styles. I retained the table styles I added as well, because the css from React Aria Table wasn't enough to simulate the layout in the wireframe.

utils/dateUtils.ts
briri: I am planning to store everything in UTC time in the database. I think that user's will want to see local time in the UI.

utils/dateUtils.ts
briri: it prompted me fix the dates in the backend 🙃

styles/_mixins.scss
fraserclark: sm, md, lg, xl, xxl, you create a more flexibility, and I think maintainable CSS system. 

It’s based on the fundamental that the viewport size rather some kind of "device", which is often ambiguous and overlapping categories of devices. We can purely focus on the dimension of width and slightly abstracting the device.

mobile screens can range from 320px to 600px, tablets from 500px to 1400px, and laptops/desktops from 1100px to 4K and beyond. There significant overlap and devices have ambiguity - Like is a 2-in-1 laptop in tablet mode a "tablet"? What about a large phone?

Also they allow helpful responsive utilities to be formed based on the same break points - classic examples are like 

Padding: py-3 for tight padding on small screens, py-md-5 for increased padding on medium screens, and py-lg-3 to reduce padding again on large screens. - This is used extensively in Bootstrap and Tailwind (and most frameworks)

Visibility:  like d-none d-md-block allow you to hide an element on smaller screens and display it on medium or larger screens. These classes become quite intuitive and means we don't need every component having its own logic etc

There are also other utitlies like flex direction, flex-row/col, css grids which often are suited to "utilities" 

In device targeting, we can use more specific heuristics if we need to figure out where we don't want hover states or whatever - but its an "add-on" etc

jupiter007: @fraserclark, I definitely agree with you that device widths are ambiguous nowadays. We can see that by looking at statcounter's width ranges for devices worldwide: https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide#monthly-202307-202407-bar

The terms used for the breakpoints is subjective, and should generally be guided by what the team finds easier to use. I am open to either sets of terms. Also, the breakpoints used in "responsive.scss" are definitely not final. My understanding is that we will be developing for `fluid layout` first, and then the breakpoints we decide to use will be less about device widths, and more about where they make sense for our design.  Based on that, it does seem like using more general terms like "`xs, sm, md, lg, xl`" would make sense. @andrewebdev and @fraserclark, if we are all in agreement, I can update the terms used in responsive.scss for now. If it becomes difficult to work with, we can revisit.

Issue #80, opened 2024-08-12
Decide and integrate on a Icon set or Library (FontAwesome or Google Material)


bofstein:
We need to pick an icon library to use.

We could use FontAwesome since it's popular. They do have some limitations when
using the non-paid icons, however.

Google’s material icons is another option. Not as expansive as the **paid**
FontAwesome set. It does have a good range of symbols and icons available for
free. Another benefit of this set is that it is consistent with other
Google products, and Android phones.

[1] - https://fontawesome.com/
[2] - https://fonts.google.com/icons

bofstein:
I'll let you all decide for sure but we probably want to lean towards the most expansive free one if it gets what we want, limit payment where we can

cazinc:
All agreed to go with Material icons @andrewebdev.

andrewebdev:
I'll add this into our base then

jupiter007:
Just a question whether we should be using svg, like the @mui/icons-material package or not.

andrewebdev:
I thought you saw my answer on the other PR related to the same question. I'll post it here again:

> I have a couple of opinions on this (and I'm happy to have them challenged :wink: )
> 
> 1. I chose the font version instead of the SVG version because it's a standard that many other platforms started adopting
> 2. We can download the font and load it locally, just like the other ones you pointed out.
> 3. I normally start out loading fonts direct from google and make use of their CDN. Performance is rarely an issue, but when it becomes a problem, that's when I download the font locally.
> 4. I would rather use the font library of the icons, over another third party that wraps them. It introduces yet another dependency (and whatever other dependencies they import), and makes security more fickle down the line.
> 
> I'm happy to be challenged on these points, and even though I feel strongly about it, I'll change it if you want me to :)

jupiter007:
Hi @andrewebdev. I did see your comment in the other Remirror ticket, but wanted to add the question about whether to use svgs or load the font family here, so that we remember to make a decision before closing this ticket.

I'll try and do some more research on it and get back to you.

Issue #74, opened 2024-07-17
Create functional ReMirror editor


bofstein:
From https://github.com/CDLUC3/dmsp_frontend_prototype/issues/6, we are going to go with ReMirror, unless we find it cannot work for us and need to change.  First step is to get a full demo working with the features we need.

andrewebdev:
@briri & @jupiter007 ; Do we know yet how we are going to store annotations in the database/backend? I'm assuming this is not yet well planned out because we might want to connect those comments with specific fields and users, some of which are not yet finalized.

jupiter007:
That's a good question @andrewebdev. We haven't put any thought into where we're going to store annotations, and how we will store it. I would think we would need the following fields:

commentId
planId
userId
content
position
timestamp
parent comment id

Also, if we're expecting real-time editing, would mysql work for that? We would probably need to use a websocket, like what would be used for a chat?

andrewebdev:
I normally use a message queue like RabbitMQ, and then I have a backend task that later sync the messages into the DB.

That said, I'm starting to wonder if comments should be done at a later phase when other technologies are more fixed in place? Might be worth a 5 minute chat tomorrow?

briri:
Yes, we can discuss tomorrow in the meeting. I hadn't given it a lot of thought since we met in Edinburgh.

Originally I was think we could use a table that uses a polymorphic association pattern. I think though that it will depend a lot on how the comments are referenced within ReMirror. How does it markup the editor's HTML to embed or link the comment to a particular part of the text?

At minimum we will want this for the MVP for answers to questions on the DMP editing screens. The polymorphic pattern will still work with that and allow us to grow. So maybe something like:
- textFieldId (the id of the thing the comment is associated with `answerId` in our initial case)
- textFieldType (to start, we will just have `Answer` here)
- commentorId (user's id)
- comment
- any markup ReMirror needs to contextualize the comment within the answer

andrewebdev:
In the richtext fields, remirror saves the "marks" for annotations in a separate JSON structure. But we can serialize this into the HTML when saving it. So In theory we can save it in the HTML. Since it's a separate JSON structure, we could also save this to the database separately. (hopefully I understood your question correctly ;) )

briri:
Ahh ok. If it can be easily serialized/deserialized from the HTML then maybe the most efficient thing to do would be to store it inline with the text. We could then just make sure we strip it out when needed like when generating a PDF export.

[...]

PR #107, opened 2024-08-30
AE - ReMirror and Material Symbols integration

andrewebdev:
## Description

- Added Material Symbols and a custom `DmpIcon` component that will render a specific icon using it's short-code. [#80]
- Added the initial version of our custom ReMirror text editor [#74].
- Also fixed a typo in a filename for the typeahead component.

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?
Tested this components by adding them to the styleguide, and manually testing functionality.

## Checklist:
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have completed manual or automated accessibility testing for my changes


andrewebdev:
I'll add some tests to the editor tomorrow. Primary goal for automated testing here is to ensure that the DmpEditor effects and events update correctly. I tested this part manually when developing, but automated tests should test that this works in future updates.

components/Editor/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  return (
+    <Group aria-label="Table Tools">
+      <MenuTrigger>
+        <Button aria-label="Menu">
+          <DmpIcon icon="table" />
+        </Button>
+        <Popover>
+          <Menu>
+            <MenuItem onAction={() => commands.createTable()}>Insert Table</MenuItem>
```
briri, line 77: Very minor, but can we change this to "Add Table" since we're using that verb for the columns and rows

components/Editor/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+          isSelected={active.orderedList()}
+        >
+          <DmpIcon icon="format_list_numbered" />
+        </ToggleButton>
+      </Group>
+
+      <Separator orientation="vertical" />
+
+      <TableGroup />
+
```
briri, line 174: It looks like the AnnotationButton is missing

briri (commented):
Thanks @andrewebdev I like the way you've organized things and it looks like it will be easy to implement the editor. 

I don't see the annotation button though when I run this locally. 
Screenshot 2024-08-30 at 8 23 31 AM

components/Editor/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+interface DmpEditorProps {
+  content: string;
+  setContent: (newContent: string) => void;
+}
+
+export function DmpEditor({content, setContent}: DmpEditorProps) {
+  const { manager, state, setState } = useRemirror({
+    extensions: () => [
+      new BoldExtension(),
```
jupiter007, line 188: L188 - @andrewebdev I see errors that at least one argument needs to be passed to BoldExtension, BulletListExtension, TableExtension, and AnnotationExtension. You can probably just pass an empty object, like "{}"

components/Editor/index.tsx
jupiter007: Yea, I don't see it either.

components/Editor/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+  return (
+    <Button onPress={handleAnnotation}>
+      <DmpIcon icon="message" />
+    </Button>
+  )
+}
+
+
+const TableGroup: React.FC = () => {
+  const chain = useChainedCommands();
```
jupiter007, line 66: L66, L67 - @andrewebdev - it looks like `chain`, `helpers`, and `getState` are not being used. Is there a reason to keep it here?

components/Icons/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+
+interface IconProps {
+  name: string;
```
jupiter007, line 5: L5 - @andrewebdev could you change `name` to `icon` so that we don't have a type def error on L8?

app/layout.tsx
```diff
@@ -27,6 +27,17 @@ export default function RootLayout({
 
   return (
     <html lang="en">
+      <head>
+        <link
```
jupiter007, line 31: @andrewebdev - I'm curious about whether it would be more performant to use svgs instead of loading an external resource. I was reading about `@mui/icons-material` [here ](https://mui.com/material-ui/icons/)and wondering if that would be a good option.

components/Editor/index.tsx
andrewebdev: Sorry, I thought we weren't doing those but rather will be building comments as a thing related to whole questions, for now, and only adding annotations later on when we know how we will reference them in the database. So to keep things neat I just removed the button for now. Apologies for the misunderstanding!

I can put it back in if you need me to?

components/Editor/index.tsx
```diff
@@ -0,0 +1,230 @@
+import React, { useEffect } from 'react';
+
+import {
+  Remirror,
+  useRemirror,
+  useRemirrorContext,
+  EditorComponent,
+  EditorState,
```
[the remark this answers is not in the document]
andrewebdev: You are right. I wonder why it didn't throw an import error like usual? I'll update it in any case, but I'm wondering why that's the case.

components/Editor/index.tsx
andrewebdev: I'll make those changes, but not sure why they were not flagged in my linter.
I run `npx eslint '**/*.ts'` to check for things like this, but it didn't flag anything.

components/Editor/index.tsx
andrewebdev: Nope, those are leftovers from development. Will remove them. Thx for catching!

components/Icons/index.tsx
andrewebdev: Yup, sorry I missed that one. For some reason the linter isn't working for me. I'll sort it when I have a moment.

app/layout.tsx
andrewebdev: I have a couple of opinions on this (and I'm happy to have them challenged :wink: )

1. I chose the font version instead of the SVG version because it's a standard that many other platforms started adopting
2. We can download the font and load it locally, just like the other ones you pointed out.
3. I normally start out loading fonts direct from google and make use of their CDN. Performance is rarely an issue, but when it becomes a problem, that's when I download the font locally.
4. I would rather use the font library of the icons, over another third party that wraps them. It introduces yet another dependency (and whatever other dependencies they import), and makes security more fickle down the line.

I'm happy to be challenged on these points, and even though I feel strongly about it, I'll change it if you want me to :)

components/Editor/index.tsx
jupiter007: @andrewebdev  - I see. Yes, so this Rich Text Editor can be used for the comments at a question level, whereas the annotation is very different in that it will be associated with inline comments. As long as we can use the Rich Text Editor example for the same commenting functionality as one our existing site, as our MVP, I think that's ok.

components/Editor/index.tsx
jupiter007: @andrewebdev this is a Typescript bug that was highlighted in VSCode during my review of your code.

app/layout.tsx
jupiter007: @andrewebdev let me do more reading on this, but I always thought using a handful of svgs was better than loading the whole font family.

jupiter007: @andrewebdev I've read some interesting articles on svgs vs icon fonts. In general, it seems like more developers are leaning towards svgs because:
1) loading icon fonts could result in some unexpected flashing
2) svgs are more accessible
3) When a visitor downloads your icon font, they are downloading every icon , even though you only want to use a handful of them
4) visual quality for svgs are better than icon fonts due to anti-aliasing in icon fonts
5) icon fonts are more difficult to style and position

One drawback of using svgs though is that inline svgs cannot be cached, but external ones can.

Here are links to some articles I read:
https://www.keycdn.com/blog/icon-fonts-vs-svgs
https://www.irigoyen.dev/blog/2021/02/17/stop-using-icon-fonts/
https://getdevdone.com/blog/icon-fonts-vs-svg-icons.html

Let me know your thoughts

andrewebdev: Okay some are valid points. I've not noticed any issues with using icon fonts in the past, but I do load these things differently in my projects. The anti-aliasing is the most compelling argument. I've not run into that myself, but I can see how that could be an issue on some systems.

I still don't think it's worth importing an entire third party library just for icons, so I'll rather download the google material icons (they are available as SVG's) and I'll use that in our `dmpicon` component.

jupiter007: @andrewebdev that sounds great. Thanks Andre.

jupiter007 (approved):

PR #152, opened 2024-10-24
Styleguide trying to unify spacing, font sizes with some global standards

fraserclark:
## Description
Added the font sizes, spacing and general clean up of the global.scss.

There is probably an on-going thing we need to keep adjusting.

I know I will go back to my current work and merge this in to apply the new spacing and font size variables where possible.

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
I have tested it in the browser across devices. I tried to make sure to add all existing variables in to maintain backwards compatibility with what exists but slowly we should replace these "legacy" variables


app/styleguide/page.tsx
```diff
@@ -51,22 +50,18 @@ import {
   CardMutedText,
 } from "@/components/Card/card";
 
-import {
-  Example,
-  BrandColor,
-  handleDelete,
-} from "./sg-components";
+import {BrandColor, Example, handleDelete,} from "./sg-components";
```
andrewebdev, line 53: @fraserclark is your editor doing automatic formatting on save? It's collapsing this import line that was deliberately spread over multiple lines.

app/styleguide/page.tsx
```diff
@@ -520,14 +994,17 @@ function Page() {
                 find specific
                 properties, and easier to add &amp; remove. Eg.
 
-                <div><pre><code>
+                <div>
+                  <pre>
+                  <code>
                   {`<SomeElement
```
andrewebdev, line 1000: @fraserclark Did you check if this is not broken now? I remember consciously formatting this without the indentation because it broke the layout in the final rendered HTML.

app/styleguide/page.tsx
```diff
@@ -538,14 +1015,14 @@ function Page() {
 
                 <div><pre><code>
                   {`import {
```
andrewebdev, line 1017: Same comment as above. The lack of indentation was intentional. Does it still render correctly on the page?

app/styleguide/page.tsx
```diff
@@ -560,16 +1037,16 @@ function Page() {
 
                 <div><pre><code>
                   {`.grid-view {
```
andrewebdev, line 1039: Same Indentation comment

andrewebdev (changes requested):
Looks good to me in general.
My only observation is that the auto-formatter in your editor is very opinionated, which results in very large diff for small, unimportant changes.
Notably, there are blocks which I intentionally created without indentation, due to layout issues on the actual rendered page. The auto-formatter decided it needed indentation, and this might have broken the rendered output. Could you verify that it's still rendering correctly?

app/styleguide/page.tsx
jupiter007: @fraserclark and @andrewebdev we should probably decide on some standard formatting and make sure all of our formatters are configured the same way. Are we all using VSCode? Maybe we can share a configuration file or something.

fraserclark: I'm using PhpStorm - because im old and use it daily! I tried VS code a few times including on DMP phase 1 stuff - but just couldn't quite get use to it so back to my old familar.

Yeah PhpStorm does a lot of auto-formatting out the box but we can set up VS Code and PhpStorm both work with .prettierrc, eslintrc and editorconfig so can probably sync all 3 update.

I'd defer to you guys on setting up the "standards" for these 3 files (Think we already 2 of the 3), I have no strong opinions on any formatting stuff.

andrewebdev: I use a combination of Emacs and NeoVim, so no common config. This is where EditorConfig is helpful. But I would always advocate for **less** auto-formatting. Editor config takes care of important stuff, like making sure everyone uses the same line-endings. But outside that auto-formatters makes too many assumptions about the code and often gets in the way, like for example the code blocks mentioned in my other comments.

jupiter007 (approved):
LGTM. The only issue was the formatting that Andre brought up.

Issue #110, opened 2024-09-11
Create right side slider component


jupiter007:
Several pages in the wireframes plan to use this "slider" functionality.

Create the shared component. See example wireframe below

<img width="859" alt="image" src="https://github.com/user-attachments/assets/72369d76-1fc4-4d65-9b39-1c9207d9180d">

briri:
Build a slider with neutral content at this point. To support all device widths

briri:
refer to notes for dialog from phase 1 for potential accessibility issues

andrewebdev:
Couple of things that still needs addressed:

- [ ] Focus the slider content when opened (Accessibility)

PR #156, opened 2024-10-25
Common Layout Containers, and Composite with Sidebar

### LINKED ISSUE #154: Build common layout containers
We need some common layout containers. These provide consistent layout styles for any components that inherit from them. These containers can also provide common hooks that could be handy for any other container, like responsive hooks etc.

The initial containers that we want is:

- [x] Base Layout Container

This is the basic layout container. We can use this directly if we don't have any fancy layout, but if we want to build custom layout containers (like the sidebar container), then those layout containers should inherit from this one.

This container does not take other components like text or paragraphs etc. It's immediate children should be other layout containers, or content containers.

- [x] Content Container

This is a container that is meant to carry the text content and other related content.

- [x] Toolbar Container

A toolbar that can take a horizontal list of items, like buttonsl. The toolbar should have properties that allow you to set it to absolute or fixed positioning if needed.

- [x] Sidebar Container

This is the first of the more complex composite containers. This container has a sidebar with related content. Direction on the container can be specified to have the sidebar left or right. The sidebar has a property that allows us to show/hide the sidebar when on larger screens. On mobile views, the sidebar will stack either above or below the main content, depending on the direction (left = above, right = below)

andrewebdev:
## Description

This PR have the new common layout containers, these serve as a base that other layout containers can be built from.

The `LayoutWithSidebar` is a composite component that demonstrates the use of the base containers.

Fixes #154 and #110 

## Type of change

## How Has This Been Tested?
I tested this locally on the Styleguide. There are sections in the styleguide where the components are documented with examples. Where applicable, I have toggle buttons to demonstrate certain dynamic functionality.

Finally, I also wrapped the entire styleguide page content with the new `LayoutWithSidebar` component, and prevented this sidebar from being closed, as an example of how it can be used in this way.

## Checklist:
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I updated the CHANGELOG.md and added documentation if necessary
- [ ] I have completed manual or automated accessibility testing for my changes


app/styleguide/page.tsx
```diff
@@ -329,17 +341,29 @@ function Page() {
...
+          <h3>Contents</h3>
 
-      <div id="sgLayout">
-        <div id="sgNav">
           <a href="#_intro">Introduction</a>
           <a href="#_brand">Branding & Colours</a>
           <a href="#_typography">Typography</a>
           <a href="#_icons">Icons</a>
           <a href="#_layout">Layout</a>
+          <a href="#_containers">Containers</a>
```
jupiter007, line 366: L366: Remember to add "id=_containers" to the Containers section so that this links to something

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
               </li>
             </ul>
 
-            <h2>Layout</h2>
-            <p>TBD ...</p>
+            <h2><code>Layout Container</code> (Base)</h2>
+            <p>The stanard <code>{`<LayoutContainer>`}</code> wraps content containers to provide
+            some common container within the layout container.</p>
+            <LayoutContainer style={{'--layout-container-background': 'var(--slate-200)'}}>
+              <ContentContainer>
```
jupiter007, line 1080: Are we planning to make our page backgrounds gray, like in the wireframes? If so, then ContentContainer should probably be given a white background. Also, I think it would be nice if we added a border-radius with a little bit of box shadow.

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+
+              <dt><code>--layout-container-background</code></dt>
+              <dd>
+                Set the container background, defaults to: <code>transparent</code>
+              </dd>
+            </dl>
+
+            <h2><code>Content Container</code> (Base)</h2>
+            <ContentContainer style={{'border': '1px solid var(--slate-600)'}}>
+              <p>This conteint is wrapped in the following container.</p>
```
jupiter007, line 1105: L1105: typo - "conteint" should be "content"

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+              <p><strong>Note</strong> that the style property is not required
+                to use the container. It's only here to show that you can use it
+                this way, and to show the outline here in the styleguide.</p>
+              <div><pre><code>{`<ContentContainer style={{'border': '1px solid var(--purple-600)'}}>
```
jupiter007, line 1109: L1109: Ideally we will not be using inline styles, and placing all styles in corresponding CSS Modules in the same directory.

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+              </dd>
+
+              <dt><code>--layout-container-background</code></dt>
+              <dd>
+                Set the container background, defaults to: <code>transparent</code>
+              </dd>
+            </dl>
+
+            <h2><code>ToolbarContainer</code> (LayoutContainer)</h2>
+            <p><code>LayoutContainer > ToolbarContainer</code></p>
```
jupiter007, line 1129: L1129: Can you change this line "<p><code>LayoutContainer > ToolbarContainer</code></p>" to use "&gt;" because it is throwing a linting error. So final one should look like "<p><code>LayoutContainer &gt; ToolbarContainer</code></p>"

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+            <ToolbarContainer style={{'--layout-container-background': 'var(--slate-200)'}}>
+              <p>Some toolbar text, maybe a Title</p>
+              <Button>Button</Button>
+            </ToolbarContainer>
+
+            <h3>CSS Variables</h3>
+            <p>Since this inherrits from <code>LayoutContainer</code> those CSS
```
jupiter007, line 1136: 1136: typo - please change "inherrits" to "inherits"

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+            <dl>
+              <dt><code>--layout-gap</code></dt>
+              <dd>
+                This container uses flexbox, and this variable sets
+                the <code>gap</code> css property. Defaults to <code>1em</code>
+              </dd>
+            </dl>
+
+            <h2>Sidebar Container (Composite)</h2>
+            <p>This is a composite layout component than contains a sidebar
```
jupiter007, line 1147: 1147: typo "than" should be "that"

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+              variables also works. In addition to those, the following is also available:</p>
+            <dl>
+              <dt><code>--layout-gap</code></dt>
+              <dd>
+                This container uses flexbox, and this variable sets
+                the <code>gap</code> css property. Defaults to <code>1em</code>
+              </dd>
+            </dl>
+
+            <h2>Sidebar Container (Composite)</h2>
```
jupiter007, line 1146: Nice sidebar!

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+            <p>This is a composite layout component than contains a sidebar
+              alongside a content container.</p>
+            <p><code>LayoutContainer > LayoutWithSidebar</code></p>
+            <p><code>ContentContainer > SidebarContainer</code></p>
+            <ToolbarContainer>
+              <Button onPress={toggleSidebarDirection}>Toggle Direction</Button>
```
jupiter007, line 1152: I noticed that on smaller mobile devices, the sidebar appears at the bottom of the page, and the toggling doesn't appear to be working.

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+  <SidebarContainer> ... </SidebarContainer>
+  <ContentContainer> ... </ContentContainer>
+</LayoutWithSidebar>
+`}
+                </code></pre></div>
+              </ContentContainer>
+            </LayoutWithSidebar>
+
+            <p><code>LayoutWithSidebar</code> can have a
+              <code>ToolbarContainer</code> in addition to the sidebar.</p>
```
jupiter007, line 1175: L1175: please add a space in front of "ToolbarContainer" and one after the word "outside on line 1177.

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+  <SidebarContainer> ... </SidebarContainer>
+  <ContentContainer> ... </ContentContainer>
+</LayoutWithSidebar>
+`}
+                </code></pre></div>
+              </ContentContainer>
+            </LayoutWithSidebar>
+
+            <h3>CSS Variables</h3>
+            <p>Since this layout inherrits from <code>LayoutContainer</code>
```
jupiter007, line 1205: L1205: typo - please change "inherrits" to "inherits"

components/Container/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  /**
+   * A handler that can notify components higher up the stack that the
+   * container changed state in some way.
+   */
+  onChange,
+}) => {
+  const thisRef = useRef(null);
+
+  useEffect(() => {
+    if (thisRef) {
```
jupiter007, line 128: L128: Could you add a check for "if (thisRef && thisRef.current) {"

components/Container/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+   */
+  sidebarPosition = 'right',
+
+  /**
+   * A handler that can notify components higher up the stack that the
+   * container changed state in some way.
+   */
+  onChange,
+}) => {
+  const thisRef = useRef(null);
```
jupiter007, line 125: L125: Remember to use Type definitions. Could you change this line to "const thisRef = useRef<HTMLDivElement>(null);".

components/Container/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+        className={`layout-container layout-with-sidebar direction-${sidebarPosition} ${className}`}
+      >
+        {children}
+      </div>
+    </>
+  )
+}
+
+interface SidebarContainerProps extends ContentContainerProps {
+  isOpen: Bool;
```
jupiter007, line 152: L152: Could you change "Bool" to "Boolean"?

components/Container/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+   * This is the direction on screen where the sidebar is located,
+   * and the direction where it will slide out of view.
+   */
+  sidebarPosition = 'right',
+
+  /**
+   * A handler that can notify components higher up the stack that the
+   * container changed state in some way.
+   */
+  onChange,
```
jupiter007, line 123: L123: I don't see "onChange" being used anywhere.

hooks/responsive.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { useState, useEffect } from 'react';
+
```
jupiter007, line 2: I'm seeing some type definition errors on this page. Would it be possible to add type definitions?

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,161 @@ function Page() {
...
+                </code></pre></div>
               </li>
             </ul>
 
-            <h2>Layout</h2>
-            <p>TBD ...</p>
+            <h2><code>Layout Container</code> (Base)</h2>
+            <p>The stanard <code>{`<LayoutContainer>`}</code> wraps content containers to provide
+            some common container within the layout container.</p>
+            <LayoutContainer style={{'--layout-container-background': 'var(--slate-200)'}}>
```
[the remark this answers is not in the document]
andrewebdev: It's a css variable, It is used extensively in the container styles.

app/styleguide/page.tsx
andrewebdev: I'll add those that style updates as you mention. But, on the styleguide I'm still going to deliberately override the backgrounds because it's only to demonstrate the different layout regions on the sidebar.

We should probably also decide on a style for "toolbars"

andrewebdev: @jupiter007 it's not really feasible to add the border and shadow, even the background color globally to `ContentContainer`. This is because of the white area in which they can be used, and the content they may contain. The primary goal for content containers is to hold content for a layout.

We can customize each content container differently with the CSS variables, on the pages where they need specific styling. ContentContainers only ensures that padding and positioning works correctly within other layouts

app/styleguide/page.tsx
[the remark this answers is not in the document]
andrewebdev: I realised why you got the typescript error for this variable, it's because typescripts default type for the style property doesn't allow for custom css variables. I changed the type def for it and it should now be okay ;)

app/styleguide/page.tsx
andrewebdev: @jupiter007 The only reason I have the inline styles here is just to show the CSS variables in use directly on the styleguide, and so that I can quickly change the colours in-place. I wouldn't use inline styles myself, but it's a quick, handy override in non-critical places like the Styleguide.

app/styleguide/page.tsx
andrewebdev: It's possible there was a misunderstanding somewhere, but I thought the component I was building, was for those pages with related content on the sidebar, and when we discussed that, people mentioned that on mobile the sidebar collapses below the content.

So this is what I've done here. I added the additional condition that if the direction is left, and it collapses, then it goes to the top instead of bottom.

Is this not what was intended? Because after our last discussion, I came to the conclusion I did the wrong thing originally, which is why I changed it.

jupiter007: Ok. I thought you were building the "slider" component. We had talked about just using "Display:flex" for the right sidebar so that it will wrap below the content. In fact, I implemented the right sidebar in the Profile page that way.

But the "slider" component will slide in additional content, rather than always displaying it.

jupiter007:
Hi @andrewebdev I confirmed that we do need a separate slider component that is separate from regular right column content which always shows. I think the regular right column content will be pretty simple, using something like "flex" to just wrap the content below the main content.

andrewebdev:
@jupiter007 , @fraserclark I've completed the updates can we possibly arrange a review meetup so I can run through the changes and elements, before you do a code review?

jupiter007:
Thanks @andrewebdev. I'll take a look first thing Friday.

jupiter007:
> @jupiter007 , @fraserclark I've completed the updates can we possibly arrange a review meetup so I can run through the changes and elements, before you do a code review?

Hi @andrewebdev. Let's plan on going over it on Tuesday along with the other PRs.

components/Container/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * This is the base layout component. Our custom layout components should
+ * inherrit from this one. So that we have some common functionality for all
+ * layout containers.
+ */
+interface LayoutContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+  style?: CustomCSSProperties,
+}
+
+export const LayoutContainer: React.FC<LayoutContainerProps> = ({
```
jupiter007, line 40: Not a big deal, but it seems like it would be easier to find these components, if they were in their own files, with them all living in the "Container" directory?

components/Container/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+  className = "",
+}) => {
+  const thisRef = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    if (thisRef) {
+      if (thisRef.current) {
+        let direction: DirectionType = null;
+
+        children?.forEach((rNode) => {
```
jupiter007, line 129: "children" prop is not always an array. You should probably add a check before using "forEach". Also, use a type guard check that "rNode" is a valid React element:
`	if (React.isValidElement(rNode) && rNode.type && typeof rNode.type === 'function') { const { name } = rNode.type as React.FunctionComponent;`

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,189 @@ function Page() {
...
+                </code></pre></div>
               </li>
             </ul>
+          </div>
+
+          <div id="_containers">
+            <h2><code>Layout Container</code> (Base)</h2>
+            <p>The standard <code>{`<LayoutContainer>`}</code> wraps content containers to provide
+            some common container within the layout container.</p>
+            <LayoutContainer>
```
jupiter007, line 1081: Will ContentContainer always live inside "LayoutContainer"? If so, we could update LayoutContainer to always include "ContentContainer".

components/Container/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+ */
+interface ContentContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+  style?: CustomCSSProperties,
+}
+
+export const ContentContainer: React.FC<ContentContainerProps> = ({
+  children,
+  id = '',
+  className = '',
+  style,
```
jupiter007, line 72: Will styles for ContentContainer ever vary, or will they always be consistent across pages?

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,189 @@ function Page() {
...
+              </ContentContainer>
+
+              <SidebarContainer>
+                <p>This is the sidebar for the slider component</p>
+              </SidebarContainer>
+            </LayoutWithSidebar>
+
+            <p>Example with a drawer instead of a sidebar</p>
+
+            <ToolbarContainer>
```
jupiter007, line 1209: When I tab around the sidebar drawer opens because of the "Close" button. Is that intentional? If not, you can use "excludeFromTabOrder" to prevent that from happening.

app/styleguide/page.tsx
```diff
@@ -1037,23 +1059,189 @@ function Page() {
...
+              <Button onPress={toggleDrawer}>Toggle Drawer</Button>
```
jupiter007, line 1210: When I open the sidebar, I'm not seeing the sidebar get focus. And when I close the sidebar, should the focus go back to where the user was last focused on the main page?

components/Container/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+  isOpen = false,
+  onClose,
+}) => {
+
+  const drawerRef = useRef<HTMLDivElement | null>(null);
+  const size = useResponsive();
+  const [isMobile, setIsMobile] = useState<Boolean>(true);
+  const [prevFocus, setPrevFocus] = useState<Element | null>(null);
+
+  useEffect(() => {
```
jupiter007, line 219: L219: This can be simplified to: 
`useEffect(() => { setIsMobile(size.viewport[0] < getSizeByName('md')[0]); }, [size]);`

components/Container/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+    if (size.viewport[0] < getSizeByName('md')[0]) {
+      setIsMobile(true);
+    } else {
+      setIsMobile(false);
+    }
+  }, [size]);
+
+  useEffect(() => {
```
jupiter007, line 227: I still feel that the "DrawerContainer" should always include a "Close" button to guarantee that there will be a way for the user to close it.

jupiter007:
@andrewebdev thanks for the updates for the sidebar and drawer. They look good. I just had a few comments on your PR.

components/Container/index.tsx
andrewebdev: You are probably right. It's been in the back of my mind, but I was more concerned with just getting things done. Will pull these into their own files when I get the other changes done.

components/Container/index.tsx
andrewebdev: I removed the "style" property. As @jupiter007 rightly points out we don't really need to change this property ever.

app/styleguide/page.tsx
andrewebdev: I'm not getting this behaviour. Are you able to replicate it with the most recent update, and if so, write steps on how to reproduce?

app/styleguide/page.tsx
andrewebdev: Unlike the DrawerPanel, the SidebarPanel don't change focus when the SidebarPanel state changes. This is due to how sidebar panels are used. I'm open to change this though, but this probably deserves a tiny bit of discussion together with @fraserclark .

DrawerPanel on the other hand do receive focus when opened, and focus goes back to previous focus when closed.

app/styleguide/page.tsx
andrewebdev: I did consider this, but decided against it because there is always a possibility that we might have a `LayoutContainer` that have multiple `ContentContainer`.

We could for example do something like this:
```html
<LayoutContainer>
  <ToolbarContainer>...</ToolbarContainer>
  <ContentContainer className="profile-form"> ... </ContentContainer>
  <ContentContainer className="contributors-form">... </ContentContainer>
</LayoutContainer>
```

components/Container/index.tsx
jupiter007: Thanks Andre

app/styleguide/page.tsx
jupiter007: @andrewebdev, I was able to reproduce by highlighting the text "Example with a drawer instead of a sidebar", and then tabbing from there. Here is a recording. The drawer is opened just from me tabbing. I didn't click the "Toggle Drawer".

https://github.com/user-attachments/assets/63a9d064-675c-4930-80a9-76d838a6378e

app/styleguide/page.tsx
```diff
@@ -329,17 +342,28 @@ function Page() {
 
   const [editorContent, setEditorContent] = useState(richtextDefault);
 
+  const [sidebarOpen, setSidebarOpen] = useState<Boolean>(true);
```
jupiter007, line 345: L345: @andrewebdev `Boolean` should actually be lower case, `boolean`.

app/styleguide/page.tsx
```diff
@@ -329,17 +342,28 @@ function Page() {
...
+  const toggleSidebar = () => {
+    setSidebarOpen(!sidebarOpen);
+  }
+
+  const [drawerOpen, setDrawerOpen] = useState<Boolean>(false);
```
jupiter007, line 350: L350: @andrewebdev `Boolean` should actually be lower case, `boolean`.

jupiter007:
Hi @andrewebdev. Do you have all your updates checked in? If so, I'll take another look before approving. Also, it looks like you need to merge the latest changes from `development` into your branch.

andrewebdev:
> Hi @andrewebdev. Do you have all your updates checked in? If so, I'll take another look before approving. Also, it looks like you need to merge the latest changes from `development` into your branch.

Just need to fix that last bug with the weird tab behaviour. I'll ping you as soon as that's checked in.

jupiter007:
@andrewebdev thanks for making all the updates. The Drawer appears to be working as expected now. However, I tried to run `npm run build` and it encountered some errors. It's always a good idea to run that command, because it catches some issues that people might miss.

I'll try and add some comments where I saw some errors.

components/Container/ContentContainer.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+
+import React from 'react';
+
+
+/**
+ * ContentContainer is the base component that is meant to hold content
+ * inside a LayoutContainer.
+ */
+interface ContentContainerProps extends React.HTMLAttributes<HTMLDivElement> {
```
jupiter007, line 10: L10: @andrewebdev  Looks like this interface is used in LayoutWithPanel, so you should probably export from here and import it there.

components/Container/LayoutContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import React from 'react';
+
+
+/**
+ * This is the base layout component. Our custom layout components should
+ * inherrit from this one. So that we have some common functionality for all
+ * layout containers.
+ */
+interface LayoutContainerProps extends React.HTMLAttributes<HTMLDivElement> {
```
jupiter007, line 11: L11: @andrewebdev I think you need to export this interface, or you can move it to the shared `app/types` file because it is used in both LayoutWithPanel and ToolbarContainer.

components/Container/LayoutWithPanel.tsx
```diff
@@ -0,0 +1,237 @@
...
+  className = "",
+}) => {
+  const thisRef = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    if (thisRef) {
+      if (thisRef.current) {
+        let direction: DirectionType = null;
+
+        function _updateClassList(name: string) {
```
jupiter007, line 46: L46: @andrewebdev I see in error here because a "function declaration" is not allowed here, but you could convert this to a "function expression" like `const _updateClassList = (name: string) => {`

components/Container/LayoutWithPanel.tsx
```diff
@@ -0,0 +1,237 @@
...
+              break;
+            }
+          }
+        }
+
+        if (Array.isArray(children)) {
+          children?.forEach((rNode) => {
+            if (!rNode) return;
+            if (!React.isValidElement(rNode)) return;
+            _updateClassList(rNode.type.name);
```
jupiter007, line 76: L76: @andrewebdev I'm seeing errors here because `rNode.type` might be a string, and therefore, there would be no `name` field on a string.

I believe you have to check whether rNode.type is a function or an object.

components/Container/LayoutWithPanel.tsx
```diff
@@ -0,0 +1,237 @@
...
+        }, 301);
+      }
+    } else {
+      if (onClose) onClose();
+    }
+  }, [stateOpen]);
+
+  useEffect(() => {
+    const keyDownHandler = (ev) => {
+      if (ev.key == 'Escape' && stateOpen) {
```
jupiter007, line 174: L173: @andrewebdev please add type `KeyboardEvent` to "ev" param.

jupiter007 (approved):

jupiter007:
@andrewebdev I approved your PR. It looks like you addressed all the comments.

Issue #180, opened 2024-11-20
[Bug] - Fix failing test for server-side component 'confirm-email'


jupiter007:
When the backend server was not running the `confirm-email` component unit test was failing. We need to update it to properly mock the fetch calls so that it would not attempt to make actual calls to the backend.

Also, when running `npm run build`, there were errors because NextJS was attempting to generate a static page by actually fetching the data from the backend. However, when the backend server was not running, the build was failing. Need to update the code so that it does not attempt a static page generation.

PR #181, opened 2024-11-20
[Bug] - Fixed the server-side component unit test, and removed graphql-codege…

jupiter007:
## Description

When the backend server was not running the confirm-email component unit test was failing. We need to update it to properly mock the fetch calls so that it would not attempt to make actual calls to the backend.

Also, when running npm run build, there were errors because NextJS was attempting to generate a static page by actually fetching the data from the backend. However, when the backend server was not running, the build was failing. Need to update the code so that it does not attempt a static page generation.

Fixes # ([180](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/180))

## Type of change
- [x} Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
Ran `npm test`, `npm run lint` and `npm run build` when the backend server was not running.

## Checklist:
- [NA] I have completed manual or automated accessibility testing for my changes

### DIFF
--- a/app/email/confirm-email/page.tsx
+++ b/app/email/confirm-email/page.tsx
@@ -1,8 +1,10 @@
-import {redirect} from 'next/navigation';
+import { redirect } from 'next/navigation';
+
+export const dynamic = 'force-dynamic'
 
 async function verifyEmail(userId: string, token: string) {
   // Verify the email address
-  const response = await fetch(`${process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT}/verify-email`, {
+  const response = await fetch(`${process.env.NEXT_PUBLIC_SERVER_ENDPOINT}/verify-email`, {
     method: 'POST',
     headers: { 'Content-Type': 'application/json' },
     body: JSON.stringify({ userId, token }),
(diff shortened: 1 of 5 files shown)


app/email/confirm-email/page.tsx
```diff
@@ -1,8 +1,10 @@
-import {redirect} from 'next/navigation';
+import { redirect } from 'next/navigation';
+
+export const dynamic = 'force-dynamic'
 
 async function verifyEmail(userId: string, token: string) {
   // Verify the email address
-  const response = await fetch(`${process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT}/verify-email`, {
+  const response = await fetch(`${process.env.NEXT_PUBLIC_SERVER_ENDPOINT}/verify-email`, {
```
andrewebdev, line 7: Just wondering if the different environment variable is intentional. If so all looks good to me.

andrewebdev (commented):
Just a single question from my side. Otherwise all looks good to me.

app/email/confirm-email/page.tsx
jupiter007: Thanks @andrewebdev. Yea, I noticed that both of those env variables had the same value, so I switched the instances of `NEXT_PUBLIC_GRAPHQL_ENDPOINT` to use `NEXT_PUBLIC_SERVER_ENDPOINT`

andrewebdev (approved):

Issue #187, opened 2024-11-26
Hook up Create/Add Section page to backend


jupiter007:
The basic layout of the Add Section page was built already, but we need to make the page functional and hook it up to backend data.

I will plan on having a separate page to Create a new Section, as opposed to Edit an existing Section.

<img width="900" alt="image" src="https://github.com/user-attachments/assets/90bad311-47a0-4bf2-aba6-59e726365d2f">

We probably need to create a separate page for Creating a new section from scratch, since we won't have a section id

![image](https://github.com/user-attachments/assets/36e3129b-249a-4e25-bb9d-8761e414c8f6)

bofstein:
@fraserclark , @jupiter007 asked about how we were doing the add tags.  As a reminder, on this particular screen, template builders can select from a pre-defined list of tags, not add their own.  We want to make sure they can see the list of options as they select.  Here is the screen on the current tool:

![Image](https://github.com/user-attachments/assets/25b4bb11-adb8-4e52-bcd8-649bb03de609)

As you can see its just checkboxes.  If we keep to the wireframe, maybe we have something like a dropdown that appears when clicked in, like this:

https://ux.stackexchange.com/questions/136510/multi-select-box-vs-dual-list-box

I don't know how accessible that is, and it's possible that will make people think they should select a lot - we generally want them to select just a couple most relevant - but it would work.  Open to other ideas!  The important thing is for them to be able to see all the options.  I don't think we need a typeahead because the words are similar enough, and most won't know the options.  Just select is fine for this short list.

@fraserclark thoughts?  Other ideas?

bofstein:
Confirmed with Juliet we do want to keep the Logic tab since we want to be able to do conditional logic on sections (e.g. "If they said "no" to an earlier question, hide this section") but the content and functionality of that will be in a later ticket

fraserclark:
I think its probably safest to effectively re-build what you currently have. Even a "fancy" select box probably suffers the same problem, usually they kind of show the ticked at the top but work as a typehead for the other options.

A quick mock-up in figma suggest it will work fairly well.

![image](https://github.com/user-attachments/assets/40b25a7d-f4e0-4aca-ad01-ca0a1bca679e)

Also might give us some flexibility to extend this in the future, which maybe other options might not

The 2 things that have been brought up before are "mis-tagging" or lack of clarity in the tagging

[...]

PR #216, opened 2024-12-20
Created 'Create Section' page, hooked in with real data and added translations

jupiter007:
## Description

Created the "Create Section" page, which will be different from "Edit Section" page, primarily because we won't be setting a Section ID until the user successfully creates a new section.

- Created a new page at: `template/[templateId]/section/create`
- Styles with css modules and added unit test for new page
- Add translations
- Added new icon for `info` to be used next to the `tags`
- `components/Editor`
-- updated `components/Editor` to use a `skeleton` to display until the editor loads. This became more of a problem as more instances of that DMPEditor were included in the same page.
-- switched `components/Editor` to use css modules
-- added two new props, `error` and `id` to the component
- Added `AddSection` and `Tags` mutation and query

Fixes # ([187](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/187))

## Type of change

## How Has This Been Tested?
This was tested both manually and using unit tests.

## Checklist:

-------------------------------------------------------------------
## Screenshots
image
image
image


app/[locale]/template/[templateId]/section/create/page.tsx
```diff
@@ -0,0 +1,417 @@
...
+    try {
+      await addSectionMutation({
+        variables: {
+          input: {
+            templateId: Number(templateId),
+            name: sectionNameContent,
+            introduction: sectionIntroductionContent,
+            requirements: sectionRequirementsContent,
+            guidance: sectionGuidanceContent,
+            displayOrder: 1,
```
Comment (briri) [inline app/[locale]/template/[templateId]/section/create/page.tsx:191]: We probably want a way to determine the next number rather than hard coding a `1` here. When a user adds a section in the current tool, it's added at the end instead of the beginning of the list. I'm not sure if we want to continue that in the new tool, but if so we'll want a function to get that number.

components/Editor/index.tsx
```diff
@@ -223,27 +232,37 @@ export function DmpEditor({ content, setContent }: DmpEditorProps) {
     setIsMounted(true)
   }, [])
 
-  const handleChange = (newState: EditorState) => {
-    const html = prosemirrorNodeToHtml(newState.doc);
+  const handleChange = ((newState: EditorState) => {
+    //prosemirror was adding an empty <p></p> when no data was passed, so we need to remove it here
```
briri, line 236: How does Prosemirror handle adding blank lines to help separate content? I believe the TinyMCE editor uses empty paragraphs to accomplish this.

We might be better off replacing it only iy what is the entire content of the html.

briri (commented):
I just had a few minor comments and one that probably requires some brief discussion on how we want the display order to work in the new tool.

app/[locale]/template/[templateId]/section/create/page.tsx
Comment (jupiter007) [reply app/[locale]/template/[templateId]/section/create/page.tsx:191]: Thanks @briri. That's a good point. It should probably go to the bottom of the list rather than the top. I can make that update and bring it up at Tuesday's meeting to confirm.

components/Editor/index.tsx
jupiter007: @briri, good catch. Prosemirror also inserts `<p></p>` for additional blank lines. I'll plan on only replacing `<p></p>` when it is the only content in the field.

app/[locale]/template/[templateId]/section/create/page.tsx
Comment (jupiter007) [reply app/[locale]/template/[templateId]/section/create/page.tsx]: @briri I updated the code to insert the new section with the max displayOrder + 1, so that it will be rendered at the bottom of the list of sections.

components/Editor/index.tsx
jupiter007: @briri I updated the code to only filter out `<p></p>` only when that is all that exists in the content.

briri: Cool thanks @jupiter007!

briri (approved):
sorry for the delay LGTM

fraserclark (approved):

Issue #308, opened 2025-02-07
Update frontend to handle errors


briri:
Need to update the frontend to handle the updated error messages from the backend.

See [backend PR #195](https://github.com/CDLUC3/dmsp_backend_prototype/pull/195)

PR #195, opened 2024-12-02
Fixed /login 404 issue related to new [locale] directory

jupiter007:
## Description

After we added the `[locale]` directory, the `/login` page was giving a **404** error in the middleware.

We need to update the logic in the middleware to add `login` and `signup` page to the exclusion list, and remove the return response for all excluded pages.

Fixes # ([194](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/194))

## Type of change
- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
Run unit tests and manually tested

## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I updated the CHANGELOG.md and added documentation if necessary
- [NA] I have added tests that prove my fix is effective or that my feature works
- [X] I have completed manual or automated accessibility testing for my changes
- [X] New and existing unit tests pass locally with my changes
- [NA] Any dependent changes have been merged and published in downstream modules

PR #316, opened 2025-02-10
Updates for new backend error handling

briri:
Addresses #308 and related to [backend PR #195](https://github.com/CDLUC3/dmsp_backend_prototype/pull/195)

**Note the backend PR should be merged first.**

- Updated the graphQL queries/mutations to include the new errors
- Updated the Profile page and UserEmail component to make use of the new field level errors and general errors
- Update the Template and Section pages to work with field level errors and general errors
- Modified tests for UserEmail component to work with new errors

To test it out, login and go to `/account/profile` then enter an invalid email address in the "Alias Emails" section and click 'Add'. You should see an error message stating that the email is invalid.

All existing unit tests are passing.


components/UpdateEmailAddress/index.tsx
```diff
@@ -191,7 +195,7 @@ calling 'refetch()' for the user query, but that didn't work. */
         })
       } else {
         // Display other errors
-        setErrors(prevErrors => [...prevErrors, 'Error when deleting email']);
+        setErrors(prevErrors => ({ ...prevErrors, general: 'Error when deleting email' }));
```
jupiter007, line 198: I almost wonder if we should have a separate state for "generalError", and should we check first whether a general error message was returned from the backend before setting it to something like "Error when deleting email"?

components/UpdateEmailAddress/index.tsx
```diff
@@ -278,8 +280,8 @@ calling 'refetch()' for the user query, but that didn't work. */
                   type="text"
                   label={t('headingAddAliasEmail')}
                   className={`${styles.addAliasTextField} react - aria - TextField`}
-                  isInvalid={errors.length > 0}
-                  errorMessage="Please enter a valid email address"
+                  isInvalid={Object.keys(errors).length > 0}
```
jupiter007, line 283: L283: Is checking for all errors, but we should just be checking for whether `errors?.email` has a value, otherwise that field-level error message will display for other errors, like general.

jupiter007:
I think this approach makes sense to me. Thanks for sharing. I just had a few comments.

jupiter007 (approved):

PR #343, opened 2025-02-19
Bug/js update dockerfile.prod

jupiter007:
## Description

Made updates to `Dockerfile.prod` and `docker-compose.yml` to try and run production in our dev environment to speed things up:
- Made updates to get Dockerfile.prod to work. 
  - Updated Dockerfile.prod
  - Updated next.config.mjs to include output `standalone`
  - Updated `docker-compose.yml`
  - Fixed lint issues introduced with changes to field-level errors


docker-compose.yml
```diff
@@ -3,11 +3,13 @@ services:
     container_name: dmsp_frontend
     build:
       context: .
-      dockerfile: Dockerfile.dev
+      dockerfile: Dockerfile.prod
```
briri, line 6: Don't you want to leave this as development?

briri (approved):
LGTM. There are a couple of other files in here that aren't applicable but that's not a big deal.

Do you really want the docker-compose though to be prod? I think that one should stay dev.

I really like the "stage" concept in the Dockerfile. I should do the same for the backend to try and make the build a bit leaner

Feel free to merge yourself when ready

jupiter007:
> LGTM. There are a couple of other files in here that aren't applicable but that's not a big deal.
> 
> Do you really want the docker-compose though to be prod? I think that one should stay dev.
> 
> I really like the "stage" concept in the Dockerfile. I should do the same for the backend to try and make the build a bit leaner
> 
> Feel free to merge yourself when ready

@briri I temporarily changed the docker-compose file to use prod because i wanted to see if it would run locally. I switched it back.

But we do have to update builderspec.yml to use Dockerfile.prod right? Otherwise, how will it know to use prod instead of dev?

jupiter007:
Also, @briri I had to fix some linting issues in some files that were breaking the build. It was related to the recent field-level error changes. I fixed them in another branch that is in PR, but I had to fix them here so the build wouldn't break.

PR #371, opened 2025-03-04
Ae question previews

### LINKED ISSUE #224: [Template Preview] As a template builder, I can preview the questions and templates I am working on to check for issues
As a user is writing questions in a template, they have the option to Preview the question.  We need to build out this functionality.
- Should show the entered data in its current state even if not Saved yet
- Should match as close as possible to what they user sees when they get to that question
- Should allow them to enter text into text fields - could cut if that's difficult but would be really nice.  Do not save the data when leaving the preview
- Just show the one question when working on question at a time, remove the breadcrumbs to rest of template
- Eventually will need Preview functionality for full template as well

*Edit (20 Feb, 2025)*

We are currently looking at ways to show the preview, without navigating aways from the page. We currently have two ideas/approaches (see attached videos)

**Example of a regular, React Aria Component Modal**

https://github.com/user-attachments/assets/01d8e164-6830-4578-8c13-59dcfd1ca4f3

**Example of an In-Page Slider Modal**

https://github.com/user-attachments/assets/8425e7ab-ea12-48a2-ac61-d65767ebf413

andrewebdev:
## Description

Added a question preview component. The component provides a single preview
button and hidden content that form part of the preview.

When the preview button is clicked, the modal opens with the preview inside.

Fixes #224

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?
Tested manually by adding the component to the styleguide, with some complex
content for testing tab and focus behaviour.

Tested by opening and closing the preview with mouse and keyboard. I also tested
that the browser history back and forward buttons behave as expected on the
component.

Lastly I also tested that scrolling does the right thing; Ensure that the
content behind the modal does not scroll.

## Checklist:


app/[locale]/styleguide/page.tsx
```diff
@@ -75,6 +75,8 @@ import {ModalOverlayComponent} from '@/components/ModalOverlayComponent';
 import ButtonWithImage from '@/components/ButtonWithImage';
 import {useToast} from '@/context/ToastContext';
 
+import QuestionPreview from '@/components/QuestionPreview';
```
Comment (jupiter007) [inline app/[locale]/styleguide/page.tsx:78]: L385 - Remember to add a link to the `Preview` demo in the Table of Contents

components/QuestionPreview/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+
+interface QuestionPreviewProps extends React.HTMLAttributes<HTMLDivElement> {
+}
+
+
+const QuestionPreview: React.FC<QuestionPreviewProps> = ({
+  children,
+  id='',
+  className='',
```
jupiter007, line 25: L25: I don't see `className` being used anywhere in the component

components/QuestionPreview/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,146 @@
...
+    );
+
+    window.history.pushState(null, "#_modal", window.location.pathname);
+    waitFor(() => {
+      const closeBtn = screen.getByTestId('preview-close-button');
+      expect(closeBtn).toBeInTheDocument();
+    });
+
+    window.history.pushState(null, "", window.location.pathname);
+    waitFor(() => {
```
jupiter007, line 141: Please add the axe accessibility unit test:
  ```
it('should pass axe accessibility test', async () => {

    const { container } = render(
      <QuestionPreview />
    );
    await act(async () => {
      const results = await axe(container);
      expect(results).toHaveNoViolations();
    });
  });
```

File: components/QuestionPreview/__tests__/index.spec.tsx [hunk 4]
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
+
+import {
+  render,
+  screen,
+  act,
```
jupiter007, line 6: L6: `act` is not used in the unit test
L12 & L13: Also, useParams, useRouter and useTranslations are not used in this test

`useTranslations` mock is configured in `jest.config.ts` so you don't have to include it in your test. Same goes for `useToast`. Also, I started a `__mocks__/common.ts` file to add some commonly used mocks for unit tests there as well.

jupiter007 (commented):
@andrewebdev the Preview component looks great. I stepped through it with `VoiceOver` and it looked good. Thanks for making the changes.

I only have a few comments.

components/QuestionPreview/__tests__/index.spec.tsx
andrewebdev: This is actually something I wanted to chat about. I actually thought you added this as one of the tools similar to translations. Which is why I didn't add the test originally.

I added it now :+1:

andrewebdev:
@jupiter007 I addressed all the other issues. Regarding your last comment on adding the preview to template, we probably need more of a discussion. We changed how this component works from how it was originally designed (as a standalone page), so we need to revisit how we are going to do previews on Templates. It's probably worth a separate ticket.

We also need to update the question types ticket(s) (or create new tickets) to now add the preview for each question on the same page where the question is created.

jupiter007:
> @jupiter007 I addressed all the other issues. Regarding your last comment on adding the preview to template, we probably need more of a discussion. We changed how this component works from how it was originally designed (as a standalone page), so we need to revisit how we are going to do previews on Templates. It's probably worth a separate ticket.
> 
> We also need to update the question types ticket(s) (or create new tickets) to now add the preview for each question on the same page where the question is created.

Thanks for making the updates @andrewebdev. I'll add an item on our agenda for next week about applying the preview to pages.

jupiter007 (approved):

Issue #370, opened 2025-03-04
Apply the use of a URL builder helper function, since our urls are getting pretty long


jupiter007:
As Fraser mentioned in a PR:

```
Like in the laravel world, there is a helper function that you can use normally it looks like

route('project.index') => "/project"
route('project.show', { projectId: 42 }) => "/project/42"
route('project.dmp.question', { projectId: 42, dmpId: 7, questionId: 999 }) => "/project/42/dmp/7/q/999"
So that is a "named" route which expects 2 params which you name in the array basically

But open it being simpler like functions

getProjectIndex() => "/project"
getProjectURL(123) => "/project/123"
getProjectDetailsURL(123) => "/project/123/details"
getProjectDmpQuestionURL(123, 555, 99) => "/project/123/dmp/555/q/99"
```

PR #374, opened 2025-03-10
Features/global-routes-helper

fraserclark:
## Description

Ticket #370 

Introduced a new utility function routePath in utils/routes.ts to generate URL paths for named routes with parameters and query strings. This function helps in maintaining a centralized route configuration and ensures consistency across the application.

**Basic - no idea if we are going to namespace it just an example**
routePath('app.login');
Output: "/en-US/login"

**Route with Path Parameters**
routePath('projects.show', { projectId: '123' });
Output: "/en-US/projects/123"

routePath('projects.dmp.section', { projectId: '123', dmpId: '456', sectionId: '789' });
Output: "/en-US/projects/123/dmp/456/s/789"

**Route with Query Parameters**
routePath('projects.index', {}, { page: 2, limit: 10 });
Output: "/en-US/projects?page=2&limit=10"

routePath('projects.index', {}, { filters: ['active', 'draft'] });
Output: "/en-US/projects?filters[]=active&filters[]=draft"

**Example of both params and query string**
routePath('projects.members.search', { projectId: '123' }, { role: 'PI', status: 'active' });
Output: "/en-US/projects/123/members/search?role=PI&status=active"

**Different locales can be passed if needed but defaults to en-US and should still work with middleware**
routePath('projects.show', { projectId: '123' }, {}, 'pt-BR');
Output: "/pt-BR/projects/123"

Fixes # ([370](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/370))

## Type of change

## How Has This Been Tested?

I have some tests using Jest 

## Checklist:


__tests__/utils/routes.test.ts
fraserclark, line 1: Pretty basics tests at the moment - but hopefully enough to get us started

app/[locale]/projects/[projectId]/dmp/[dmpid]/page.tsx
Comment (fraserclark) [inline app/[locale]/projects/[projectId]/dmp/[dmpid]/page.tsx:1]: Can mostly ignore the changes here - but you'll see the routePath in action

utils/routes.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (!(paramName in params)) {
+      throw new Error(`Missing required parameter: ${paramName} for route ${name}`);
+    }
+    // Use a template string to create a new string instead of replace
+    path = path.split(param).join(String(params[paramName])) as typeof path;
+  }
+
+
+// Add query string if query params are provided
+  if (Object.keys(query).length > 0) {
```
fraserclark, line 99: ChatGPT can take alot of credit for this part!!

URLSearchParams() weirdly doesnt seem to work with arrays eg

tag[]=blue&tag[]=yellow

so its why its slightly hacky to manually make the query string for arrays

utils/routes.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @param params - Object containing route parameters (optional)
+ * @param query - Object containing query parameters (optional)
+ * @param locale - Optional locale override
+ * @returns Generated URL path
+ */
+export function routePath(
+  name: keyof typeof routes,
+  params: Record<string, string | number> = {},
+  query: Record<string, QueryParamValue> = {},
+  locale: string = 'en-US'  // Default locale
```
fraserclark, line 76: Locale would ideally be pulled from browser, JWT or library - I had a look at the middleware but got stuck a little so backed out for now

 @jupiter007 you might have insight into this with your work on the middleware.

jupiter007: @fraserclark , yea, we don't have to worry about the `locale`. That param and functionality can be removed. The middleware will handle it, and, also, just constructing relative urls will use the locale that is already in place. You can try it out, just return `  return `${path}`;` on line 131

utils/routes.ts
```diff
@@ -0,0 +1,137 @@
...
+  const path_pattern = routes[name];
+  if (!path_pattern) {
+    throw new Error(`Route not found: ${name}`);
+  }
+
+  // Replace parameters in the path
+  let path = path_pattern;
+  const paramNames = path.match(/:[a-zA-Z_]+/g) || [];
+
+  for (const param of paramNames) {
```
jupiter007, line 88: L88 - We can update this so that the entire string isn't reconstructed for every param. Something like the following should work in its place:
 ```
 let path = path_pattern.replace(/:([a-zA-Z_]+)/g, (_, paramName) => {
    if (!(paramName in params)) {
      throw new Error(`Missing required parameter: ${paramName} for route ${name}`);
    }
    return String(params[paramName]);
  });
```

File: utils/routes.ts [hunk 4]
```diff
@@ -0,0 +1,137 @@
...
+    const searchParams = new URLSearchParams();
+    const manualParams: string[] = [];
+
+    for (const [key, value] of Object.entries(query)) {
+      if (value === undefined) {
+        continue; // Keep `undefined` values out of the query string
+      }
+      if (Array.isArray(value)) {
+        // Manually format arrays with `[]` notation
+        value.forEach((v) => manualParams.push(`${encodeURIComponent(key)}[]=${encodeURIComponent(String(v))}`));
```
jupiter007, line 109: `URLSearchParam` uses different encoding than `encodeURIComponent` so would it work to apply `encodeURIComponent` in both scenarios?

utils/routes.ts
fraserclark: If we let the middleware always handle it, dont we end up with needless redirects? As all the URLs would be excluding the locale

utils/routes.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  // Special case for root path
+  if (path === '/') {
+    return `/${locale}`;
+  }
+
+  // Add locale prefix
+  return `/${locale}${path}`;
+}
+
```
jupiter007, line 133: I was trying to see if there's a way to see what paths the route keys correspond to when we are using them. For example, if I am typing ('routePath('projects.show'), I was hoping to see the path that it ultimately corresponds to while I am typing. 

It worked for me when I did this:
```

type RouteName = keyof typeof routes;

export function routePath<T extends RouteName>(
  name: T,
  params: Record<string, string | number> = {},
  query: Record<string, QueryParamValue> = {},
  locale: string = 'en-US'  // Default locale
): `${(typeof routes)[T]}` { 

  return path as `${(typeof routes)[T]}`;
```

jupiter007 (commented):
Thanks @fraserclark for working on this. It looks great. I only had a few comments and questions.

utils/routes.ts
jupiter007: I'm not too concerned about the redirects. But, having said that, we could also grab the current user's locale from the path within the `routePath` util itself, so that each component doesn't have to extract a locale and then send it as a param to routePath(). 

We could use something like the following, since the locale is always in the first segment
```
    const pathname = window.location.pathname;
    const segments = pathname.split('/');
```

__tests__/utils/routes.test.ts
jupiter007: @fraserclark are you ok with moving this test into the `utils/__tests__` folder? At least for me, it's easier to find the tests when they are in the same directory as the file they are testing.

fraserclark: All good moved it now

utils/routes.ts
fraserclark: Slightly different from how you have it but welcome for you to change 

But right now its producing for me this 
image

utils/routes.ts
fraserclark: I have updated to use the first segment aswell

fraserclark:
Updated PR based on changes above

utils/__tests__/routes.test.ts
fraserclark, line 1: Moved file as suggested

utils/routes.ts
```diff
@@ -0,0 +1,157 @@
...
+// Define query param value types
+type QueryParamValue =
+  string
+  | number
+  | boolean
+  | string[]
+  | number[]
+  | undefined;
+
+export const supportedLocales = ['en-US', 'pt-BR'] as const;
```
fraserclark, line 12: Placeholder for now  to have some validation on these - but probably should think about how we want to do this as middleware has similar

utils/routes.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  // Automatically detect and validate locale from URL if not provided or invalid
+  if (!locale || !supportedLocales.includes(locale as SupportedLocale)) {
+    if (typeof window !== 'undefined') {
+      const segments = window.location.pathname.split('/');
+      const potentialLocale = segments[1];
+      if (supportedLocales.includes(potentialLocale as SupportedLocale)) {
+        locale = potentialLocale;
+      } else {
+        locale = 'en-US'; // explicit fallback
```
fraserclark, line 109: Figure we don't want an error; just a push to en-US?

jupiter007: yes, thanks Fraser

PR #407, opened 2025-04-07
Updated app to use nodeJS v22

jupiter007:
## Description

Update app to use NodeJS v22, which is the Active LTS version

Fixes # ([144](https://github.com/CDLUC3/uc3sys/issues/144))

## Type of change
This is a Chore to update the nodeJS version to the current Active LTS version.

## How Has This Been Tested?
Ran all existing tests against this change.
Manually tested app and checked for any errors.

## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I updated the CHANGELOG.md and added documentation if necessary
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have completed manual or automated accessibility testing for my changes
- [X] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #415, opened 2025-04-09
Fixed wrong node image in Dockerfile.prd and fixed a number of linting issues

jupiter007:
## Description

I merged [this node version update](https://github.com/CDLUC3/dmsp_frontend_prototype/pull/407), but the Codepipeline failed because i forgot to update the node images everywhere in the `Dockerfile.prod` file.

I then found a bunch of typescript errors when trying to commit the change, so i fixed those.


briri (approved):
LGTM

Issue #437, opened 2025-04-29
While editing a Section, the breadcrumbs show "Add a new section"


bofstein:
Its a bit confusing to see "Add a new section" in the list here since as a user I know I'm just updating.  It doesn't seem like adding a new section should be part of this pathway.  Not a high priority.

![Image](https://github.com/user-attachments/assets/07417f16-6d22-4eb3-86fd-879d55b551a3)

jupiter007:
This was addressed in this PR: https://github.com/CDLUC3/dmsp_frontend_prototype/pull/438

Issue #436, opened 2025-04-29
Updating a Section moves it to the bottom of the template


bofstein:
After I make changes to a Section, it moves the order to the bottom of the template.  Order should not change when updates are made.

Order being updating "Dissemination":

![Image](https://github.com/user-attachments/assets/8494cf6d-aab6-4d4d-a9f0-4fff44ad9f2e)

Order after I updated "Dissemination" section:

![Image](https://github.com/user-attachments/assets/a458256d-32c2-4819-86f8-c41933155a4e)

jupiter007:
The frontend should be using the `displayOrder` for the sections, but it currently is just using the order that the Sections are returned in. It should be a fairly easy fix.

Issue #435, opened 2025-04-29
Saving a Section or Questions should bring the user back to the main Template page, not stay on the Edit page


bofstein:
After updating the section and clicking "Update Section," it should go back to the Template Overview page, with my changes saved.  Instead of stays on the Section page.  It's hard to get back to the Template page since there's no "Back" button and the breadcrumbs (the only way back I know) are small and note visible after clicking save.

![Image](https://github.com/user-attachments/assets/c1fecc37-b387-4f5d-af35-789b9a4c50c2)

Same for editing Questions.  In the Wireframe it was "Save and Add" to make it clear you're adding when you do that and going back to the template.  Otherwise it is hard to get back, and we'd need some sort of back/return/close button on this page.

![Image](https://github.com/user-attachments/assets/5a5fe8ea-80eb-4ff7-a536-228fa50f949f)

jupiter007:
Made update to use `Save and add` for button text in `Update section` page as well as `Edit Question` page.

made update to redirect user back to `Edit Template` page after submitting form.

jupiter007:
Changes are contained in this PR: https://github.com/CDLUC3/dmsp_frontend_prototype/pull/438

PR #438, opened 2025-04-29
[Bug] - Fixed Section ordering issue on Edit Template page

jupiter007:
## Description

Currently, the `Sections` on the `Edit Template` page, at `/templates/[templateId]` are not ordered based on their `displayOrder` value.

- The page was updated to use the `displayOrder` for `Sections`.
- Updated the `Edit Template` and `Update Section` page to use the new `routePath()`
- Updated `Update Section` page to redirect to the `Edit Template` page after submitting a change to the page
- Updated `Update Section` and `Edit Question` forms to use button that says `Save and add` per most recent requirement changes
- After forms are submitted for above, the user should be redirected to the `Edit Template` page
- Removed `Add a new section` from the breadcrumbs on the `Update Section` page

Fixes # ([436](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/436))
([435](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/435))
([437](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/437))

## Type of change

## How Has This Been Tested?
This has been manually tested, as well as testing via unit tests.

## Checklist:


app/[locale]/template/[templateId]/page.tsx
```diff
@@ -121,6 +121,11 @@ const TemplateEditPage: React.FC = () => {
     }
   );
 
+  const sortSections = (sections: Section[]) => {
+    // Create a new array with the spread operator before sorting
+    return [...sections].sort((a, b) => (a.displayOrder ?? 0) - (b.displayOrder ?? 0));
```
Comment (briri) [inline app/[locale]/template/[templateId]/page.tsx:126]: wondering if we want to default to zero. This would place the section at the top of the list.
Under what conditions would it be null/undefined? On a new section?

briri (approved):
LGTM. I just had one question about whether or not we want to default a displayOrder to zero

fraserclark (approved):
Looks good to me

app/[locale]/template/[templateId]/page.tsx
Comment (fraserclark) [reply app/[locale]/template/[templateId]/page.tsx:126]: Yeah I agree, I think we'd always want to have a display order by default, it gets complicated if its null.

To be fair, "create section"  deals with it - it gets the current display order and +1's - but might be good to set it at the database level too for robustness

https://github.com/CDLUC3/dmsp_frontend_prototype/blob/d81b891e8bb0509a282774d06c831f3bc995742e/app/%5Blocale%5D/template/%5BtemplateId%5D/section/create/page.tsx#L111C9-L111C28

Comment (briri) [reply app/[locale]/template/[templateId]/page.tsx:126]: 👍🏻 yeah, I have a [backend ticket](https://github.com/orgs/CDLUC3/projects/14/views/3?pane=issue&itemId=84553821&issue=CDLUC3%7Cdmsp_backend_prototype%7C130) to ensure the DB has the orders in place.

I think calling that `getNewDisplayOrder` function could work

Comment (jupiter007) [reply app/[locale]/template/[templateId]/page.tsx:126]: @briri and @fraserclark the problem arises because the type definition from the backend says that `displayOrder` is `string | null`. So trying to figure out what to do when it is `null` is the problem. Perhaps if the displayOrder is null, I can just add it to the end of the list?

Comment (jupiter007) [reply app/[locale]/template/[templateId]/page.tsx:126]: > 👍🏻 yeah, I have a [backend ticket](https://github.com/orgs/CDLUC3/projects/14/views/3?pane=issue&itemId=84553821&issue=CDLUC3%7Cdmsp_backend_prototype%7C130) to ensure the DB has the orders in place.
> 
> I think calling that `getNewDisplayOrder` function could work

That's great. It makes sense for the backend to deliver an ordered list.

Comment (jupiter007) [reply app/[locale]/template/[templateId]/page.tsx:126]: > 👍🏻 yeah, I have a [backend ticket](https://github.com/orgs/CDLUC3/projects/14/views/3?pane=issue&itemId=84553821&issue=CDLUC3%7Cdmsp_backend_prototype%7C130) to ensure the DB has the orders in place.
> 
> I think calling that `getNewDisplayOrder` function could work

@briri I went ahead and merged the ticket as is for now since you have a ticket to order the sections on the backend.

PR #465, opened 2025-05-06
Removed duplicate graphqlServerActionHandler.ts and serverAuthHelper.…

jupiter007:
## Description

In trying to debug why Server Actions is not working on DEV environment, I noticed there were duplicate `graphqlServerActionHandler.ts` and `serverAuthHelper` files and one set did not have the updated SERVER_ENDPOINT env variables. I'm hoping this will resolve the issue we're seeing on Dev environment.


briri (approved):
LGTM

Issue #462, opened 2025-05-06
[Feature] - Need to replace Remirror with TinyMCE for now, so that we can update important dependencies in NextJS


jupiter007:
Currently, using Remirror is preventing us from updating `next` version to v15, due to `getRootProps` errors when using `use strict`, and also preventing us from updating our `react` and `react-dom` versions because it's using older dependencies.

It is not regularly maintained as well, so we conducted a spike, and decided to switch to TinyMCE, even though it doesn't have free inline commenting: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/409

Being able to update our dependencies soon is critical.

PR #487, opened 2025-05-07
Feature/js implement tiny mce rte

jupiter007:
## Description

- Added a `TinyMCEEditor` component
- Included it in the styleguide

Once these changes are approved, we can create a separate ticket to update the pages that use the Remirror editor with the TinyMCE Editor.

Keep in mind that to have a `self-hosted  tinymce editor` I had to copy the `tinymce` folder from `node_modules` to `public/tinymce`. Not all the files are loaded on page load though - it should only load the `plugins` that we specified in the `TinyMCEEditor`.

Fixes # ([462](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/462))

## Type of change

## How Has This Been Tested?
Manually tested in styleguide, and actually plugged into a `Edit Section` page temporarily to make sure it works with existing components.

## Checklist:


components/TinyMCEEditor/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+        branding: false, // removed the tinyMCE branding
+        statusbar: false, //removes the bottom status bar
+        selector: `#${elementId}`,
+        menubar: true,
+        max_height: 400,
+        min_height: 200,
+        plugins: [
+          'autoresize',
+          'table',
+          'code'
```
briri, line 48: We don't allow the user to manipulate the underlying `<html>` markup directly. In the past, people have done it improperly and it ends up breaking pages when we attempt to render (e.g. an admin adds invalid custom markup to a question's guidance and then a researcher picks that template and tries to view the question, it breaks the layout of the page)

components/TinyMCEEditor/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      window.tinymce.remove(`#${elementId}`);
+
+      // Initialize new editor
+      window.tinymce.init({
+        license_key: 'gpl',
+        promotion: false,// Removes TinyMCE promotional link
+        branding: false, // removed the tinyMCE branding
+        statusbar: false, //removes the bottom status bar
+        selector: `#${elementId}`,
+        menubar: true,
```
briri, line 42: I don't think we want to display the menubar, most of the options don't have functionality the user would want. In the current tool we just offer functionality through the toolbar

briri (approved):
Looks good overall @jupiter007.

My comments are based on how we're currently using TinyMCE. It would be worth reviewing with @bofstein to see which features she wants us to offer.

components/TinyMCEEditor/index.tsx
jupiter007: > I don't think we want to display the menubar, most of the options don't have functionality the user would want. In the current tool we just offer functionality through the toolbar

Unfortunately, as far as I can tell, the `menubar` is needed to provide the `table` menu. Otherwise, there's no way to add a table.

components/TinyMCEEditor/index.tsx
jupiter007: > We don't allow the user to manipulate the underlying `<html>` markup directly. In the past, people have done it improperly and it ends up breaking pages when we attempt to render (e.g. an admin adds invalid custom markup to a question's guidance and then a researcher picks that template and tries to view the question, it breaks the layout of the page)

Good point. I removed everything from the `menubar` except `Edit` an `Table`

components/TinyMCEEditor/index.tsx
jupiter007: @briri actually, I just realized I can add the `Table` functionality to the `Toolbar`, so I'll get rid of the `menubar` afterall.

jupiter007: Ok Brian, I pushed up my change. Thanks for your review.

PR #524, opened 2025-05-30
Fixed issue where existing title doesn't show when editing template name/title

### LINKED ISSUE #475: Edit Template name should have existing name in there
When you click to edit Template name, the current Name should be in the text field to start with.

andrewebdev:
## Description

Fixed but where the template name is blank when editing the name for an existing
template.

To facilitate this fix I had to clean up a lot of events and effects, that were
all either redundant, causing confusion, or causing blocking effect updates.

Fixes #475

## Type of change

## How Has This Been Tested?
- I tested this manulaly by editing the title, changing it etc.
- I updated the existing tests to verify my changes.
- I also ensured that existing tests pass so that I didn’t introduce new errors.

## Checklist:

### DIFF
--- a/components/PageHeaderWithTitleChange/__tests__/index.spec.tsx
+++ b/components/PageHeaderWithTitleChange/__tests__/index.spec.tsx
@@ -100,7 +100,7 @@ describe('PageHeaderWithTitleChange', () => {
   });
 
   it('should render without crashing', () => {
-    render(<PageHeaderWithTitleChange title="Test Title" />);
+    render(<PageHeaderWithTitleChange {...defaultProps} />);
   });
 
   it('should pass axe accessibility test', async () => {


components/PageHeaderWithTitleChange/index.tsx
```diff
@@ -38,19 +34,24 @@ const PageHeaderWithTitleChange: React.FC<PageHeaderProps> = ({
...
   actions,
   className = '',
-  handleInputChange,
-  handleTitleChange,
-  newTitle = '',
+  onTitleChange,
 }: PageHeaderProps) => {
 
   const [isEditing, setIsEditing] = useState(false);
+  const [newTitle, setNewTitle] = useState<string>(title);
```
jupiter007, line 41: @andrewebdev, I feel that the `PageHeaderWithTitleChange` component should not manage the state of the `title`. That should stay with the `TemplateEditPage` so that PageHeaderWithTitleChange can remain stateless. I think it's better to have separation of concerns.

jupiter007: To solve this issue without moving the "setTitle" state to `PageHeaderWithTitleChange`, you can do the following simple change:

1) In the `TemplateEditPage` component:
- Set the `newTitle` in the useEffect
```
  // Need to set this info to update template title
  useEffect(() => {
    if (data?.template) {
      setTemplateInfoState({
        templateId: Number(data.template.id),
        name: data.template.name || '',
        visibility: data.template.visibility || null,
      });
      // Set newTitle to the template name only if newTitle is not already set
      if (!newTitle) {
        setNewTitle(data.template.name || '');
      }
    }
  }, [data]);
```
- Remove the `newTitle` prop from `PageHeaderWithTitleChange` and set `title={newTitle}`:
```
      <PageHeaderWithTitleChange
        title={newTitle}
        description={description}
        showBackButton={false}
        breadcrumbs={
          <Breadcrumbs>
            <Breadcrumb><Link href="/">{BreadCrumbs('home')}</Link></Breadcrumb>
            <Breadcrumb><Link href="/template">{BreadCrumbs('templates')}</Link></Breadcrumb>
            <Breadcrumb>{template.name}</Breadcrumb>
          </Breadcrumbs>
        }
        className="page-template-overview"
        handleTitleChange={handleTitleChange}
        handleInputChange={handleInputChange}
      />
```
2) In the `PageHeaderWithTitleChange`:
- Remove the `newTitle` prop from the component
- Change `value` in `<FormInput>` to equal `title || ''`:
```
              <FormInput
                name="templateTitle"
                type="text"
                label="Template title"
                value={title || ''}
                inputClasses="titleChange-input"
                placeholder="Enter new template title"
                onChange={handleInputChange}
              />
```
and please hard-code the `name` to something like `templateTitle`.

jupiter007: @andrewebdev yes, the initial implementation used `newTitle` and `title` which was not optimal, but the changes I suggested above is only using `title` and it keeps state management of `newTitle` in the parent `TemplateEditPage`, rather than moving it to `PageHeaderWithTitleChange`.

With your solution, you have two separate state variables tracking the same logical piece of data:
- `title` is passed to PageHeaderWithTitleChange as `template.name`. In PageHeaderWithTitleChange, you are tracking title changes with the state `const [newTitle, setNewTitle] = useState<string>(title);`. We could have syncing issues, and in general we want to keep the editing state in the parent.

jupiter007: @andrewebdev - I'm going to go ahead and approve your changes. Later, we should probably think of another way to do this so that we don't have state changes for `title` in the shared PageHeaderWithTitleChange component.

Issue #516, opened 2025-05-27
ORCID connection has wrong color and text


bofstein:
On the Connections page, the Connect ORCID button and SSO connection buttons are light grey instead of something readable (not sure if this would be blue or black here):

![Image](https://github.com/user-attachments/assets/07cbfba6-b233-4b76-8fdf-f397feee4012)

Additionally, I'm not ure why we have both Connected and Not Connected states on the page.  I could be wrong but I thought this connection was implemented for real, in which case this page should show the proper state for this account.  And should have proper text like "ORCID Connected" instead of mentioning user state.

The page also changes between ORCID and ORCiD - not sure which is correct but we should be consistent.

fraserclark:
PR in review
https://github.com/CDLUC3/dmsp_frontend_prototype/pull/525

bofstein:
This one is not fix on Dev, it still shows the wrong button color and text

![Image](https://github.com/user-attachments/assets/91928b07-38e2-4ec5-9ec8-ee2d05747639)

bofstein:
Confirmed on Dev

PR #525, opened 2025-05-30
Bug/orcid connection update 516

fraserclark:
## Description

Fixes from #516
- Fix heading size
- Fix button styling
- Fix buttons
- Made it 2 column layout for consistency
- Add translations

## Type of change

## How Has This Been Tested?
Ran the unit tests and browser tested

## Checklist:


app/[locale]/account/connections/page.tsx
```diff
@@ -4,55 +4,106 @@ import React from 'react';
...
+    isOrcidConnected
+      ? 'orcidConnectionConnected.content'
+      : 'orcidConnection.content',
+    {
+      link: (chunks) =>
+        `<a href="https://orcid.org/" target="_blank" rel="noopener noreferrer">${chunks}</a>`
+    }
+  );
+*/
   const orcidContentString = sanitizeHtml('ORCID provides a persistent identifier - an ORCID iD - that distinguishes you from other users. Learn more at <a href="https://orcid.org/" target="_blank" rel="noopener noreferrer">ORCID.org</a>.');
```
Comment (fraserclark) [inline app/[locale]/account/connections/page.tsx:45]: This was existing code - I tried to make the t() work but was having failing unit tests using t.markup. I need to do more research on how to mock this properly but for the demo I just kept it simple with what was there already.

app/[locale]/account/connections/page.tsx
```diff
@@ -4,55 +4,106 @@ import React from 'react';
...
+              ) : (
+                <ConnectionSection
+                  type='orcidconnected'
+                  title={t('orcidConnectionConnected.title')}
+                  content={orcidContentString}
+                  btnUrl='/users/auth/orcid/test'
+                  btnImageUrl='/images/orcid.svg'
+                  btnText={t('orcidConnectionConnected.btnText')}
+                />
+              )}
```
Comment (fraserclark) [inline app/[locale]/account/connections/page.tsx:85]: Kept the old code but re-named it orcidconnected and put the connection in a "if connected" which is just mocked since we dont have the backend yet.

app/[locale]/account/connections/__tests__/page.spec.tsx
```diff
@@ -24,7 +27,7 @@ describe('Connections page', () => {
   afterEach(() => {
     jest.restoreAllMocks();
   })
-
+/*
```
Comment (jupiter007) [inline app/[locale]/account/connections/__tests__/page.spec.tsx:30]: @fraserclark can you remove the commented unit test? I think it's ok since we're mocking the page header.

app/[locale]/account/connections/page.tsx
Comment (jupiter007) [reply app/[locale]/account/connections/page.tsx:45]: @fraserclark, I just asked chatGPT, and it gave me this mock for the `t.markup` function that works:
```
jest.mock('next-intl', () => {
  const t = (key: string) => key;
  t.markup = (key: string, values: Record<string, (chunks: string) => string>) => {
    // Simulate interpolation of the `link` function
    const chunks = key; // Simplified, in real case you'd simulate how the lib works
    const interpolated = Object.entries(values).reduce((acc, [name, fn]) => {
      return acc.replace(`{${name}}`, fn(chunks));
    }, key);
    return interpolated;
  };
  return {
    useTranslations: jest.fn(() => t)
  };
});
```

jupiter007 (approved):
@fraserclark thanks for the updates. It looks good to me. I just had one comment about removing some commented out code in a unit test, and provided a suggestion for mock the useTranslations so it will work with `t.markup`.

PR #560, opened 2025-06-23
Feature/question types implementation

jupiter007:
## Description

- Built out all the question type components and added components to the `components/Form/QuestionComponents` directory
- Updated the "Add Question" and "Edit Question" components to allow for the adding and editing of the new question types
- Updated QuestionView, which is used to show a Preview of the Question that is being added or edited. Added handlers to allows users to interact with the questions in the Preview mode.
- Built a "MultiSelect" component (see screenshot below). It's supposed to be an accessible approach, but I'd love any feedback on it.
-   Create the "Number" and "Currency" components using React Aria Component's `NumberField`. This allows users to increment and decrement the field using buttons.
- Updated and added unit tests

The different question type look can be seen by clicking on the `Preview` button when you are adding a new question type. I've included screenshots below.

I provided label fields for `dateRange`, `numberRange` and label and help text fields for `affiliationSearch` in the "Add Question" form because their json schemas include labels. We might want to add labels to some other question types.

Fixes # (157)

## Type of change

## How Has This Been Tested?
This has been tested manually and by using unit tests.

## Checklist:

### Radio Buttons
image

### Checkboxes
image

### Select box
image

### MultiSelect
image

### Number Field
image

### Number Range
image

### Date
image

### Date Range
image

### Boolean 
image

### Affiliation Search
image

### DIFF
--- a/components/QuestionAdd/index.tsx
+++ b/components/QuestionAdd/index.tsx
@@ -246,15 +246,6 @@ const QuestionAdd = ({
       }
       return;
     }
-
-    const merged = {
-      ...parsed,
-      attributes: {
-        ...('attributes' in parsed ? parsed.attributes : {}),
-        ...getOverrides(questionType),
-      },
-    };
-
     return {
       ...parsed,
       attributes: {


app/[locale]/template/[templateId]/q/new/page.tsx
```diff
@@ -76,30 +88,27 @@ const QuestionTypeSelectPage: React.FC = () => {
...
   const filterQuestionTypes = (
     questionTypes: QuestionTypesInterface[],
     term: string
   ): QuestionTypesInterface[] =>
-    questionTypes.filter(qt =>
-      [qt.name, qt.usageDescription].some(field =>
-        field?.toLowerCase().includes(term.toLowerCase())
-      )
-    );
+    questionTypes.filter(qt => {
```
Comment (jupiter007) [inline app/[locale]/template/[templateId]/q/new/page.tsx:104]: Update to search the `json`, as well as `name` and `usageDescription` fields for question types.

app/[locale]/template/[templateId]/q/new/page.tsx
```diff
@@ -197,7 +205,7 @@ const QuestionTypeSelectPage: React.FC = () => {
               <div>
                 {/*Show # of results with clear filter link*/}
                 {(searchTerm.length > 0 && searchButtonClicked) && (
-                  <div className={styles.searchMatchText}> {Global('messaging.resultsText', { name: filteredQuestionTypes?.length || 0 })} - <Link onPress={resetSearch} href="/" className={styles.searchMatchText}>{QuestionTypeSelect('links.clearFilter')}</Link></div>
+                  <div className={styles.searchMatchText}> {Global('messaging.resultsText', { name: filteredQuestionTypes?.length || 0 })} - <Link onPress={resetSearch} href="#" className={styles.searchMatchText}>{QuestionTypeSelect('links.clearFilter')}</Link></div>
```
Comment (jupiter007) [inline app/[locale]/template/[templateId]/q/new/page.tsx:208]: Switched from `href='/'` to `href='#"` to prevent redirection when user clicks "clearFilter" link

fraserclark:
One thing I forgot to ask was required field - Perhaps its not in the mock data but should it be etc? Will it be? 

I know there was some discussion around being truly required and "Marking it as required" but really it is optional

components/QuestionView/index.tsx
```diff
@@ -88,6 +216,167 @@ const QuestionView: React.FC<QuestionViewProps> = ({
...
+              />
+            )}
+
+          </>
+        );
+      }
+      case TEXT_FIELD_QUESTION_TYPE:
+        return (
+          <FormInput
+            name="textField"
```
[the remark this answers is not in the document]
jupiter007: Thanks @fraserclark. That's a good observation.

I updated the calls to `FormInput` to explicitly pass the `max` and `min` values, and `pattern`.

When working on this ticket, I had the same questions about adding `label`, and other info to the `json` schemas. Currently, if you look at the `json` schemas in `questionTypes`, you'll see that `label` isn't included in most of the schemas except for `dateRange`, `numberRange` and `typeaheadSearch`. I even added a comment in a backend bug I created, that mentions adding that and other info in the saved `json`: https://github.com/CDLUC3/dmsp_backend_prototype/issues/322

I think it's a good thing to discuss maybe in our next Dev meeting.

jupiter007:
> One thing I forgot to ask was required field - Perhaps its not in the mock data but should it be etc? Will it be?
> 
> I know there was some discussion around being truly required and "Marking it as required" but really it is optional

We have not yet updated the backend `json` schemas with `required`. We can discuss this at our Dev meeting.

components/QuestionAdd/hooks/useAddQuestion.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  if (typeof source === 'string') {
+    try {
+      return JSON.parse(source);
+    } catch (err) {
+      logECS('error', 'getParsedQuestionJSON error', {
+        error: err,
+        url: { path: routePath('template.q.slug') }
+      });
+      return null;
```
andrewebdev, line 45: I mentioned this in my previous comment. I think a function should either fail outright by throwing and exception, so that you can catch and handle that on the outside, rather than returning null.

Also, logECS is called here (in two places in fact), and then the same logECS function is called outside in the other code. This will result in two logs for one error.

My changes mentioned earlier should address some of this, but it might also be worth looking at the Custom Error I added which has the logging built in. By just throwing that error (or subclass of it), you will automatically get the logging, as well a named exception you can deliberately catch later.

see: `LoggedError` inside `utils/exceptions.ts`

components/QuestionTypeCard/index.tsx
```diff
@@ -17,7 +17,7 @@ const QuestionTypeCard: React.FC<QuestionTypeCardProps> = ({ questionType, handl
 
   const json = JSON.parse(questionType.json);
   const questionTypeId = json.type;
-  const qType = JSON.stringify(questionTypeId)?.trim().replace(/^"|"$/g, '');//Since the question type was taken from the JSON string, it could have quotes or slashes that need to be removed.
+  let qType = JSON.stringify(questionTypeId)?.trim().replace(/^"|"$/g, '');//Since the question type was taken from the JSON string, it could have quotes or slashes that need to be removed.
```
andrewebdev, line 20: Just wondering the reason this changed to `let` instead of `const`? It doesn't look like the qType value changes elsewhere in the component.

styles/_elements.scss
```diff
@@ -647,9 +647,9 @@
     margin-left: 0;
   }
   
-  .checkbox-group {
-    justify-self: end; /* Aligns checkbox group to the right */
-  }
+  // .checkbox-group {
+  //   justify-self: end; /* Aligns checkbox group to the right */
+  // }
```
andrewebdev, line 652: Maybe just delete this outright. In my experience if we move on from this ticket people will just forget about this code and one day wonder why it's sitting here. It's a small enough bit of code that can easily be re-created. So in my opinion just put it in the bin ;)

Alternatively, if there is a reason to keep it around, we should at least have a comment stating why it's commented out and remains here.

andrewebdev (commented):
@jupiter007 that's my review pass done as well. Just a couple of notes from my side. I don't think they are urgent, so if you feel we can move on without addressing them, then just let me know in slack, and I'm happy to approve.

app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
```diff
@@ -135,38 +161,92 @@ const QuestionEdit = () => {
...
+  };
+
+  // Handler for typeahead help text changes
+  const handleTypeAheadHelpTextChange = (value: string) => {
+    setTypeAheadHelpText(value);
+
+    if (parsedQuestionJSON && (parsedQuestionJSON?.type === "typeaheadSearch")) {
+      const updatedParsed = structuredClone(parsedQuestionJSON); // To avoid mutating state directly
+
+      if (updatedParsed.graphQL &&
```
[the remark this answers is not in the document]
Comment (jupiter007) [reply app/[locale]/template/[templateId]/q/[q_slug]/page.tsx]: @andrewebdev good catch. I removed the redundancy

app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
```diff
@@ -190,65 +270,160 @@ const QuestionEdit = () => {
...
-    if (selectedQuestion) {
+    if (selectedQuestion?.question) {
+      const q = selectedQuestion.question;
 
-      const q = selectedQuestion?.question || null;
-      const json = getParsedQuestionJSON(q);
-      const questionType = json.type ?? questionTypeIdQueryParam;
-      const questionTypeFriendlyName = Global(`questionTypes.${questionType}`) || '';
+      try {
+        const parsed = getParsedQuestionJSON(q);
```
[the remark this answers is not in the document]
Comment (jupiter007) [reply app/[locale]/template/[templateId]/q/[q_slug]/page.tsx]: @andrewebdev you have a good point, although I didn't want to wrap try-catch around everything when I used the getParsedQuestionJSON. But, I updated it to pass back an error that can be displayed, and I made sure to check whether `parsed` is null when I use it.

app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
```diff
@@ -300,56 +467,75 @@ const QuestionEdit = () => {
...
                   <div className={styles.optionsWrapper}>
                     <p
-                      className={styles.optionsDescription}>{QuestionEdit('helpText.questionOptions', { questionType })}</p>
+                      className={styles.optionsDescription}>{t('helpText.questionOptions', { questionType })}</p>
                     <QuestionOptionsComponent
                       rows={rows}
-                      setRows={setRows}
-                      questionType={questionType}
+                      setRows={updateRows}
+                      questionJSON={question ? getParsedQuestionJSON(question) : {}}
```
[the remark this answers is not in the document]
Comment (jupiter007) [reply app/[locale]/template/[templateId]/q/[q_slug]/page.tsx]: @andrewebdev good catch. I updated this code.

components/QuestionAdd/hooks/useAddQuestion.ts
jupiter007: @andrewebdev I updated `useAddQuestion` to use the shared `getParsedQuestionJSON` which handles errors now.

styles/_elements.scss
jupiter007: @andrewebdev I deleted the unused style

jupiter007:
@andrewebdev Thanks for your Preview. I addressed all your comments.

Please take another look, and if you are ok with the changes, please approve.

jupiter007:
> @jupiter007 yup, that looks good.
> 
> I like the explicit error handling you have in the updated get parsed question json function. (I personally prefer errors as values, over exceptions, the way that Go, Rust and Zig handle them).
> 
> Another pattern to consider for future code of this nature, is to use Promises. The error handling can then happen in the catch resolver.
> 
> Either way, I've approved these latest changes.

Thanks for your review @andrewebdev

Issue #548, opened 2025-06-10
Improve unit test suite to flag critical errors as not passing


bofstein:
As discussed in rebuild sync meeting, if there are critical warnings that need to be addressed, the test show show this result and ideally not pass, rather than rely on the person to scroll up and see that there are warnings and decide if they're critical or not.  We should look into configuring the tests so that they don't pass if there's something that needs to be addressed.

jupiter007:
I think we can do something like:
```
const originalConsoleError = console.error;
const originalConsoleWarn = console.warn;beforeEach(() => {
  jest.restoreAllMocks(); // reset any mocks between tests

  console.error = (...args) => {
    throw new Error(`Unexpected console.error: ${args.join(' ')}`);
  };

  console.warn = (...args) => {
    throw new Error(`Unexpected console.warn: ${args.join(' ')}`);
  };

  // Fail test on uncaught promise rejections
  process.on('unhandledRejection', (reason) => {
    throw new Error(`Unhandled Promise rejection: ${reason}`);
  });

  // Fail test on unhandled exceptions
  process.on('uncaughtException', (err) => {
    throw err;
  });
});

afterEach(() => {
  // Restore original console behavior
  console.error = originalConsoleError;
  console.warn = originalConsoleWarn;

  process.removeAllListeners('unhandledRejection');
  process.removeAllListeners('uncaughtException');
});
```
to our jest.setup.ts file and it will catch unhandled exceptions and, consequently, fail the test.

However, if might be too strict and pick up on exceptions we intentially throw for our unit tests.

PR #565, opened 2025-06-30
Chore/js improve unit testing to capture extraneous errors even with passing tests

jupiter007:
## Description

Currently, we may get passing unit tests in NextJS, but we sometimes get exceptions that are not captured because we are not testing for them in the unit test, and jest/react-testing-library just lets those errors go through. These unhandled errors could disrupt unit test results by giving false positives. These errors also contribute to a lot of noise in the logs when we run our tests.

Therefore, I added another test in `jest.setup.ts` that will hopefully capture those errors. I found that if I filter the exceptions on `Error:`, it seems to just catch real errors, rather than catching all `console.error` that includes ones we log intentionally in our unit tests.

This new test already caught one of those issues with the unit test in graphqlHelper.ts. I added `navigateion.ts` util so that we can adequately test the redirects for INTERNAL_SERVER_ERROR, and address the navigation error.

Fixes # ([548](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/548))

## Type of change

- [X] Chore

## How Has This Been Tested?
Ran unit tests

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-### Updated
-- Clean up connections page and buttons [#516]
+### Added
+- Added a test in `jest.setup.ts` to catch extraneous errors that are not tested or in a unit test.
 
 ### Updated
+- Clean up connections page and buttons [#516]
 - Added spacing on the `Account Profile` for the demo [#509]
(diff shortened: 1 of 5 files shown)


briri (approved):
LGTM!

Issue #564, opened 2025-06-27
Need to remove "Sample text" field from the "Add Question" page when question types don't need it


jupiter007:
My understanding is that the "Sample text" field should only be displayed for the `textArea` questionType. Currently, it's also display on the "Add Question" form for the following question types:

	- Text field
	- Number field
	- Number range
	- Currency field
	- Email field
	- Url field
	- Boolean or Yes/No field
	- Date field
	- Date Range
	
We should remove it for those question types.

PR #573, opened 2025-07-01
Removed the Sample Text field for most question types

jupiter007:
## Description

Removed `Sample text` field from the "Add Question" and "Edit Question" forms for all question types except for `textArea`.

Fixes # ([564](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/564))

## Type of change

## How Has This Been Tested?
Unit tests

## Checklist:

### DIFF
--- a/app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
+++ b/app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
@@ -60,11 +60,11 @@ import {
   QuestionTypesInterface
 } from '@/app/types';
 import {
-  TEXT_QUESTION_TYPE,
   RANGE_QUESTION_TYPE,
   TYPEAHEAD_QUESTION_TYPE,
   DATE_RANGE_QUESTION_TYPE,
-  NUMBER_RANGE_QUESTION_TYPE
+  NUMBER_RANGE_QUESTION_TYPE,
+  TEXT_AREA_QUESTION_TYPE
 } from '@/lib/constants';
 import {
   isOptionsType,
@@ -628,7 +628,7 @@ const QuestionEdit = () => {
 
                 />
 
-                {!hasOptions && (
+                {questionType === TEXT_AREA_QUESTION_TYPE && (
                   <FormTextArea
                     name="sample_text"
                     isRequired={false}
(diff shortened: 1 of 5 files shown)


app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
```diff
@@ -60,11 +60,11 @@ import {
   QuestionTypesInterface
 } from '@/app/types';
 import {
-  TEXT_QUESTION_TYPE,
```
Comment (andrewebdev) [inline app/[locale]/template/[templateId]/q/[q_slug]/page.tsx:63]: Is this removed because we are just using `TEXT_AREA...` instead? (I don't recall :sweat_smile: )

components/QuestionAdd/__tests__/index.spec.tsx
```diff
@@ -120,20 +110,23 @@ describe("QuestionAdd", () => {
...
       meta: {
+        asRichText: true,
         schemaVersion: "1.0"
       },
-      type: "text",
+      type: "textArea",
       attributes: {
         pattern: null,
+        rows: null,
+        cols: null,
```
andrewebdev, line 120: Just wondering, and this is a bit of a technical point. But is it work using rows and columns? It's attributes on the field that came before CSS, and doesn't have any purpose these days when we can just resize the text area with CSS.

andrewebdev (approved):
Looks good, just two minor comments from me. They are not important, just thoughts I had while going through the code, so I'm approving.

app/[locale]/template/[templateId]/q/[q_slug]/page.tsx
Comment (jupiter007) [reply app/[locale]/template/[templateId]/q/[q_slug]/page.tsx:63]: @andrewebdev it's because for certain question types, like `text`, `number`, `date`, there is currently nothing that the user creating the question needs to add, like a `label`, `help text`, etc. The current `json` schemas for those question types don't support it yet. So I removed it because we didn't need it.

components/QuestionAdd/__tests__/index.spec.tsx
jupiter007: @andrewebdev yea I'm not sure how much good it's serving us, but decided to leave it for now. We can always remove it later if we don't need it.

jupiter007:
@andrewebdev thanks for your review

Issue #578, opened 2025-07-08
Create "expand/collapse" functional for right sidebar for Best Practices


jupiter007:
Update the right sidebar on pages like `/projects/[projectId]/dmp/[dmpid]/s/[sid]` and `/projects/[projectId]/dmp/[dmpid]/s/[sid]/q/[qid]`, as well as preview pages to display the Best practices content so that they are collapsed, and can be expanded.

Update the above pages to use the new `Best Practices` functionality

bofstein:
@jupiter007 where can I see this?  On a page like https://dmptool-dev.uc3dev.cdlib.net/en-US/projects/5/dmp/14/s/357 I still see the best practice links without being able to collapse it.  I then tried to start a question to see it and got a 404 so maybe I picked a bad template?

bofstein:
To clarify, this is just Expand and Collapse for the text, correct?  That part looks good to me.  You can't collapse the whole Sidebar I think we eventually want but that may be a different ticket.

jupiter007:
Yes @bofstein the sidebar is currently not collapsible, unless it is the "Comments" or "View Sample Answer" drawers on the Question Details page

PR #586, opened 2025-07-10
[Feature] - Added a new component to expand and collapse content

jupiter007:
## Description

- Added a new `ExpandableContentSection` component that allows use to `expand` and `collapse` content, especially for the `Best Practices` right sidebar, but can be used for any content
-- Added a `.md` file for usage of the component
-- Added unit test for new component
- Updated `QuestionView` to use the new `ExpandableContentSection` for the `Best Practices` right sidebar

Fixes # ([578](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/578))

## Type of change

## How Has This Been Tested?
This was tested manually and with a new unit test

## Checklist:

## Important Note:

## Screen recording

https://github.com/user-attachments/assets/01cb6721-1984-4ef5-9201-fbef1b4a95d6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - fix translation string to use the correct tags for the `Account Profile` [#515]
 
 ### Fixed
+- Fixed broken links on the `Project overview` page that go to the individual plans [#575]
 - Removed the `Sample text` field from the Question Add/Edit forms for all question types except for `textArea` [#564]
 
 
--- a/app/[locale]/projects/[projectId]/dmp/[dmpid]/__mocks__/planQueryMock.ts
+++ b/app/[locale]/projects/[projectId]/dmp/[dmpid]/__mocks__/planQueryMock.ts
@@ -5,7 +5,8 @@ export const mockPlanData = {
       template: {
         id: 2,
         name: "General"
-      }
+      },
+      name: "NSF-CISE: Computer and Information Science and Engineering"
     },
     fundings: [
       {
--- a/app/[locale]/projects/[projectId]/dmp/[dmpid]/__tests__/page.spec.tsx
+++ b/app/[locale]/projects/[projectId]/dmp/[dmpid]/__tests__/page.spec.tsx
@@ -117,7 +117,7 @@ describe('PlanOverviewPage', () => {
   it('should render plan data', async () => {
     render(<PlanOverviewPage />);
 
-    expect(screen.getByRole('heading', { name: 'Reef Havens: Exploring the Role of Reef Ecosystems in Sustaining Eel Populations' })).toBeInTheDocument();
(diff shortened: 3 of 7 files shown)


components/QuestionView/index.tsx
```diff
@@ -8,6 +8,7 @@ import { CalendarDate, DateValue } from "@internationalized/date";
 
 import {
   Button,
+  Link
```
andrewebdev, line 11: This import doesn't seem to be used elsewhere in the file.

components/QuestionView/index.tsx
```diff
@@ -47,6 +50,8 @@ import {
   AffiliationSearchQuestionComponent,
   BooleanQuestionComponent
 } from '@/components/Form/QuestionComponents';
+
+//
```
andrewebdev, line 54: Empty comment

components/ExpandableContentSection/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  // Truncate the text up to the max length (i.e., specified summaryCharLimit)
+  const truncateText = (text: string, maxLength: number): { truncated: string; wasTruncated: boolean } => {
+
+    const truncated = text.substring(0, maxLength);
+    const lastSpace = truncated.lastIndexOf(' ');
+    const finalText = lastSpace > 0 ? truncated.substring(0, lastSpace) : truncated;
+
+    // Add ellipsis to summary or truncated text
+    return { truncated: finalText + '...', wasTruncated: true };
```
andrewebdev, line 53: Just thinking out loud if we should use the actual unicode character for ellipse? `…` instead of thhree periods.

andrewebdev (commented):
Looks good overall, just two small comment.

components/QuestionView/index.tsx
jupiter007: Thanks. I removed it. 

I implemented a fix in another branch to allow `linting` to work again so it can catch stuff like this :)

components/QuestionView/index.tsx
jupiter007: Thanks Andre. I added a comment there

sfisher (approved):
Looks good and thanks for adding the component.  I don't think I have anything to add over the existing review (plus its been a while since I wrote any React).

andrewebdev (approved):

components/ExpandableContentSection/index.tsx
jupiter007: > I'm just a bit of a typographic snob, that's all @jupiter007 :) I dabble in a bit of writing in my spare time so my editors are set up to automatically use. Anyway you can ignore it if you want. I was just thinking out loud and wondering what you thoughts would be on that.

No worries @andrewebdev. It's a good question. I just don't like the way that the unicode looks and I think users are more familiar with the `...`

PR #612, opened 2025-07-17
Add popular funders to the funder search page when creating a new project

### LINKED ISSUE #380: Adding "Most popular" funders to the Funding Search page
Currently, the page at `/projects/create-project/funding-search` has hard-coded "Most popular funders" below the search box (see below).
But there is no backend data that specifies which funders are the most popular.

We could look at our data possibly to figure that out, and then update the Funding page with that info.

The static funding page has been built, and the search functionality is being worked on here: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/335

But this ticket is to follow up later and decide what info we may want to include below the search box. And what do we want to do if the search returns no results?

Image

andrewebdev:
## Description

Added the most popular funders to the funding search page _when creating a new project_.

Fixes #380

## Type of change

## How Has This Been Tested?

- Created new automated test and related mocks to check that the correct graphQL queries are called.
- Automated tests also check that selecting a popular funder will add the funder to the project just like searching for a funder would.
- Did manual testing where I clicked through to this page, verified that the popular funders are show, and that I can click to pick one of the popular funders.

## Checklist:

## Important Note:

### DIFF
--- a/app/[locale]/projects/[projectId]/funding-search/page.tsx
+++ b/app/[locale]/projects/[projectId]/funding-search/page.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { useState, useRef } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { useTranslations } from 'next-intl';
 import { useParams, useRouter } from 'next/navigation';
 import { routePath } from '@/utils/routes';
@@ -9,6 +9,9 @@ import { LoggedError } from "@/utils/exceptions";
 
 import {
   AffiliationSearch,
+  // usePopularFundersQuery,
+  FunderPopularityResult,
+  usePopularFundersLazyQuery,
   useAddProjectFundingMutation,
   ProjectFundingErrors,
 } from '@/generated/graphql';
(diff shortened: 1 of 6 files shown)


andrewebdev:
Thank you for the approval @jupiter007 . Could you possible do a quick sanity check on the code since I pushed an update earlier to get 100% code coverage.

app/[locale]/projects/[projectId]/funding-search/page.tsx
```diff
@@ -69,7 +68,8 @@ const CreateProjectSearchFunder = () => {
    * @param {ProjectFunderErrors} errs - The errors from the graphql response
    */
   function checkErrors(errs: ProjectFundingErrors): string[] {
-    if (!errs) return [];
+    const noErrors = Object.values(errs).every(val => val === null);
```
Comment (jupiter007) [inline app/[locale]/projects/[projectId]/funding-search/page.tsx:71]: L71: Do we need this early return here? It seems a bit redundant since the loop below that does a check for null and returns an empty array if there are no valid errors.

Comment (jupiter007) [reply app/[locale]/projects/[projectId]/funding-search/page.tsx:71]: It seems like it could be updated to: 
```
function checkErrors(errs: ProjectFundingErrors): string[] {
  const typedKeys: (keyof ProjectFundingErrors)[] = [
    "affiliationId",
    "general",
    "projectId",
    "status",
  ];

  return typedKeys.reduce((acc: string[], key) => {
    const val = errs[key];
    if (val) acc.push(val);
    return acc;
  }, []);
}
```

app/[locale]/projects/[projectId]/funding-search/page.tsx
```diff
@@ -126,23 +126,25 @@ const CreateProjectSearchFunder = () => {
   };
 
   function onResults(results: FunderSearchResults, isNew: boolean) {
-    if (results) {
-      const items = (results.items ?? [])
-        .filter((f): f is AffiliationSearch => f != null);
+    let validResults: AffiliationSearch[];
+    if (results.items && results.items.length > 0) {
```
Comment (jupiter007) [inline app/[locale]/projects/[projectId]/funding-search/page.tsx:130]: Good catch adding the `length` check.

jupiter007 (approved):
Great job in getting the coverage to 100%. That's awesome.

I found one bug in the process of re-reviewing where an `aria-label` value is being passed incorrectly. Also, just noted a couple of minor possible improvements to the code. But otherwise, it looks good. I tested manually as well.

app/[locale]/projects/[projectId]/funding-search/page.tsx
Comment (jupiter007) [reply app/[locale]/projects/[projectId]/funding-search/page.tsx:71]: Ok, makes sense @andrewebdev. Just leave it as is then.

Issue #686, opened 2025-08-05
[Bug] Picking a Template for a new Plan is only returning 20 results


bofstein:
When picking a Template to start a plan, it only loads 20 which makes it hard to find specific Templates to use if they don't appear.  We should have the full list, with potential pagination(?) available

<img width="1125" height="740" alt="Image" src="https://github.com/user-attachments/assets/4c3b0e62-a8fb-4216-987a-74b64874b1c8" />

jupiter007:
Put into PR review.

## Expected functionality
- Expected functionality is that if the project has funder, and there are published templates that match those funders, then the page will render with `funder checkboxes` initially checked, and the list of templates filtered to only show those funder templates
- If there are no funders, but the list of templates returned contain a `best practice` template, then a `Best practice` checkbox will be display when the page first renders, and the list of templates will be filtered to display only the `best practice` templates.
- If the project has no funders and no best practice templates, then no checkboxes will be displayed.
### Pagination
- Pagination can be tested on the page by clicking either the "Next" or "Previous" buttons to advance or reverse the pages, or users can click directly on a page. Ellipsis were used to prevent too many pages from showing at once.


## To Test
- Make sure you use the backend branch `feature/JS-paginationOptions-update-to-include-additionalData-378` with this branch
- You can add funders to your projects by clicking on the "Funders" link on the `Projects Overview` page and then clicking on `Add funders`
- To remove funders, you can go into the `projectFundings` db table and remove it from there
- To set templates as `bestPractice`, you can set the `bestPractice` field to `1` in the `versionedTemplates` table

- [ ] Test with projects that have funders
- [ ] Test with projects without funders
- [ ] Test with and without best practice templates
- [ ] Verify pagination works with different result set sizes
- [ ] Confirm filter combinations work as expected

## Screen recordings
### When project has `funders`


https://github.com/user-attachments/assets/512a9f69-67c9-425c-8e07-6473b03cac0b


### When project has no `funders` but has `best practice` templates

https://github.com/user-attachments/assets/349272da-b6a2-49ac-b91e-3d5023fb42c0

### When project has no `funders` and no `best practice` templates

https://github.com/user-attachments/assets/59c7f60b-3bf3-4d93-aa6b-8ef3f1ebe317

bofstein:
[...]

PR #693, opened 2025-08-10
Updated Plan Create page to use new pagination queries and added Pagination component

jupiter007:
## Description

This ticket was filed because the page never loaded more than the default 20 results that the backend started to return when it updated to use pagination queries.

- I updated the `Plan Create` page to use the new `pagination queries`, moving off of the manual `Load more` functionality we had in place
- Added use of `routePath` on page
- Added a new, accessible `Pagination` component, which is used in the updated `Plan Create` page. When adding the pagination, I tried to comply with the WCAG 2.5.5 `AA` tap size requirements of `24px x 24px`
- Updated unit test for `Plan Create` to use the `MockedProvider` and added some new unit tests
- Added translation keys

Fixes # ([686](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/686))

## Type of change

## How Has This Been Tested?
This was tested manually, and with unit tests

## Checklist:
- [*] Any dependent changes have been merged and published in downstream modules

*** NOTE**: Since we switched to paginated data, the frontend only gets data in increments, but it needed to know whether the full set of data contained a `bestPractice` template, and also it needed the list of all the unique affiliations that would be in the total response, so that it would know whether to show the `affiliation/funder` checkboxes or `bestPractice` checkbox when the page was rendered.

The PR for the backend change is here: https://github.com/CDLUC3/dmsp_backend_prototype/pull/379

The backend branch that I updated is `feature/JS-paginationOptions-update-to-include-additionalData-378`
 
## Important Note:

## Business requirements
- Expected functionality is that if the project has funder, and there are published templates that match those funders, then the page will render with `funder checkboxes` initially checked, and the list of templates filtered to only show those funder templates
- If there are no funders, but the list of templates returned contain a `best practice` template, then a `Best practice` checkbox will be display when the page first renders, and the list of templates will be filtered to display only the `best practice` templates.
- If the project has no funders and no best practice templates, then no checkboxes will be displayed.
### Pagination
- Pagination can be tested on the page by clicking either the "Next" or "Previous" buttons to advance or reverse the pages, or users can click directly on a page. Ellipsis were used to prevent too many pages from showing at once.

## To Test
- Make sure you use the backend branch `feature/JS-paginationOptions-update-to-include-additionalData-378` with this branch
- You can add funders to your projects by clicking on the "Funders" link on the `Projects Overview` page and then clicking on `Add funders`
- To remove funders, you can go into the `projectFundings` db table and remove it from there
- To set templates as `bestPractice`, you can set the `bestPractice` field to `1` in the `versionedTemplates` table

- [ ] Test with projects that have funders
- [ ] Test with projects without funders
- [ ] Test with and without best practice templates
- [ ] Verify pagination works with different result set sizes
- [ ] Confirm filter combinations work as expected

### When project has `funders`

https://github.com/user-attachments/assets/512a9f69-67c9-425c-8e07-6473b03cac0b

### When project has no `funders` but has `best practice` templates

https://github.com/user-attachments/assets/349272da-b6a2-49ac-b91e-3d5023fb42c0

### When project has no `funders` and no `best practice` templates

https://github.com/user-attachments/assets/59c7f60b-3bf3-4d93-aa6b-8ef3f1ebe317

### DIFF
--- a/components/Pagination/index.tsx
+++ b/components/Pagination/index.tsx
@@ -1,0 +1,39 @@
+'use client';
+import React from 'react';
+
+interface PaginationProps {
+  currentPage: number;
+  totalPages: number;
+  hasPreviousPage: boolean | null;
+  hasNextPage: boolean | null;
+  fetchTemplates: (page: number) => Promise<void>
+}
+
+export default function Pagination({ currentPage, totalPages, hasPreviousPage, hasNextPage, fetchTemplates }: PaginationProps) {
+  return (
+    <div>
+      <button onClick={() => fetchTemplates(currentPage - 1)} disabled={!hasPreviousPage}>
+        Prev
+      </button>
+
+      {[...Array(totalPages)].map((_, index) => {
+        const page = index + 1;
+        return (
+          <button
+            key={page}
+            data-page={page}
+            data-current={currentPage}
+            onClick={() => fetchTemplates(page)}
+            disabled={page === currentPage}
+          >
+            {page}
+          </button>
+        );
+      })}
+
+      <button onClick={() => fetchTemplates(currentPage + 1)} disabled={!hasNextPage}>
+        Next
+      </button>
+    </div>
+  );
+}
--- a/components/Pagination/index.tsx
+++ b/components/Pagination/index.tsx
@@ -6,41 +6,36 @@ interface PaginationProps {
   totalPages: number;
   hasPreviousPage: boolean | null;
   hasNextPage: boolean | null;
-  bestPractice?: boolean;
-  selectedOwnerURIs?: string[];
-  fetchTemplates: (args: {
-    page?: number;
-    bestPractice?: boolean;
-    selectedOwnerURIs?: string[]
-  }) => Promise<void>;
-
+  handlePageClick: (page: number) => void;
 }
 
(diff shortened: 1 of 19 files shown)


components/Pagination/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      pages.push(totalPages);
+    }
+
+    return pages;
+  };
+
+  const pagesToRender = getPages();
+
+  return (
+    <nav aria-label="pagination" className={styles.pagination}>
```
jupiter007, line 57: Made sure to make pagination accessible by using `<nav>` wrapper and `aria-labels`

styles/globals.scss
```diff
@@ -248,3 +248,17 @@ body {
     top: 0;
   }
 }
+
+.clear-filter {
```
jupiter007, line 252: Adding the `clear filter` link styling in shared location, since we will be applying it in other places

jupiter007:
> I think this all looks good. The only small suggestion I’d make is that we might not need pagination at the top of the page, we kind of want people to review the list? Id say just at the bottom of the list is fine
> 
> That said, if others feel differently, I’m happy to go with whatever

Thanks @fraserclark. That's a good point about people reviewing the list. That makes sense. I put pagination at the top originally because I wanted to make sure that people knew that there were more items, but I guess they will find that out when they scroll down :)

PR #694, opened 2025-08-11
(#489) Updated the CSS to allow wrapping tags

### LINKED ISSUE #489: [Bug] - The tag text extends outside of the parent container when the text is too long
On mobile, the tag text extends outside of the container when too long.

This can, actually, also happen for desktop, because currently it's using a `white-space: nowrap`. We need to think about other solutions. Especially since we can have varying widths due to different languages, etc.

Image

andrewebdev:
## Description

Allow for tags in the checkbox group to wrap when the screen size is small.
Also added a bit better alignment of the info buttons on the smaller sizes.

Fixes #489

## Type of change

## How Has This Been Tested?

- Only did manual testing in the browser, resizing the view to different widths
  to see the final result.

## Checklist:
- [ ] I updated the CHANGELOG.md and added documentation if necessary

## Important Note:

### DIFF
--- a/styles/form/_checkbox.scss
+++ b/styles/form/_checkbox.scss
@@ -23,7 +23,7 @@
     display: flex;
     align-items: center;
     justify-content: center;
-    margin-top: 0.3rem;
+    margin-top: 0.2rem;
   }
 
   svg {
@@ -77,7 +77,6 @@
 
   .react-aria-Checkbox {
     margin-bottom: 0.5rem;
-
   }
 
   label {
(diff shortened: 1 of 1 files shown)


styles/form/_checkbox.scss
```diff
@@ -124,30 +123,39 @@
...
-    align-items: center;
+    align-items: start;
     gap: 0.25rem;
+
+    @media (max-width: 768px) {
+      white-space: wrap;
+    }
   }
 
+  @media (max-width: 460px) {
```
jupiter007, line 137: Same here. I think we want to use `device` rather than `@media`. Or at least I believe that was the plan when we created the `mixin`: 
```@mixin device($breakpoint) {
  @if map.has-key($breakpoints, $breakpoint) {
    @media screen and (min-width: map.get($breakpoints, $breakpoint)) {
      @content;
    }
  } @else {
    @warn "Unfortunately, no value could be retrieved from `#{$breakpoint}`. "
          + "Available breakpoints are: #{map.keys($breakpoints)}.";
  }
}
```

jupiter007:
> @jupiter007 that final change we discussed just pushed now. If you want to have a look and give me the okay, I can merge it.

Thanks @andrewebdev. I'll take a look.

Issue #760, opened 2025-09-05
Add text on Search Funder page in Project Start


bofstein:
Let's add text here that says "Select the funder to use for this project.  This can be the funder you plan to apply to or where you already have funding from.  You'll be able to add additional funders later if you have more than one."

<img width="1570" height="843" alt="Image" src="https://github.com/user-attachments/assets/c8330430-ceac-4535-9e1a-09284eab2703" />

PR #793, opened 2025-09-09
(#760) Added descriptive text to funding-search page

andrewebdev:
## Description

Added descriptive text to the funding-search page when creating a new project.

Fixes [#760](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/760)

## Type of change

- [X] Chore

## How Has This Been Tested?
- Ran the existing unit tests to make sure they still pass
- Manual testing by navigating to the page and checking that the text exist.

## Checklist:
- [ ] I have performed a self-review of my own code


app/[locale]/projects/[projectId]/funding-search/page.tsx
```diff
@@ -150,7 +150,7 @@ const CreateProjectSearchFunder = () => {
     <>
       <PageHeader
         title={trans('headerTitle')}
-        description=""
+        description={trans('pageIntro')}
```
Comment (jupiter007) [inline app/[locale]/projects/[projectId]/funding-search/page.tsx:153]: L153: I believe this is the wrong translation key. Don't you need `headerDescription`?

Comment (andrewebdev) [reply app/[locale]/projects/[projectId]/funding-search/page.tsx:153]: :facepalm: yes, you are right. I changed the key last minute to fit more with the how other keys are set up and I forgot to change it here. Good catch. Will fix

Comment (andrewebdev) [reply app/[locale]/projects/[projectId]/funding-search/page.tsx]: Should now be fixed.

jupiter007 (approved):
Thanks Andre. LGTM.

PR #795, opened 2025-09-09
Static: Admin Org templates

fraserclark:
## Description

Duplicated the logged in user templates to add Organization Templates page.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore

## How Has This Been Tested?
- unit tests and browser tests

## Checklist:

## Important Note:


fraserclark:
Hey this is kind of a discussion point @andrewebdev @jupiter007 

We have dashboards for logged-in users (eg Templates, Projects). We also need admin dashboards that are right now 99% the same UI, but with a broader GraphQL scope. I have duplicated the dashboard page into an admin path so that admins see organisation-wide data. 

All click-throughs id imagine will go to the same manage/edit pages, which enforce authorisation to show/hide bits.

So the approach
- Keep a distinct Next.js route for admin, eg:
  - /templates/page.tsx        → user view
  - /admin/templates/page.tsx  → admin view
- Reuse the same presentational components and editor pages. Only the page-level query and any admin-specific filters differ.

Pros for why duplicate the page
- Clear URL space: an explicit /admin path is easy to protect, test, and reason about. It also helps with analytics, redirects, and policies/auth.
- Safer data scope: separate files reduce the risk of leaking admin data to user views due to a missed conditional?
- Likely divergence: organisation dashboards may need different filters and defaults over time. Separate pages let us evolve them without stacking role conditionals.
- Perhaps slightly simpler code: fewer branches per page and cleaner dependency chains. Shared UI stays in components rather than conditional logic - the page as it stands is quite component-based anyway.
- Perhaps easier to test the data scope?

Trade-offs
- There is page-level duplication to keep in sync which could be annoying
- Mitigation: keep all shared UI and fragments in common components/modules. Only the query and wiring live in the two pages. 

So this is kind of a question as I have ~3 other dashboards to do that are admin sectioned - so do we agree to this duplicate dashboard but leave the pages being conditional?

Should I proceed with this or would you prefer conditional queries in the standard "logged in user" files - I would then add the "static elements" to these instead and just put in some fake conditionals

fraserclark:
Replaced it with the paginated query and updated the tests based on the /templates directory

jupiter007 (approved):
Thanks for making that change @fraserclark. Looks good to me.

PR #803, opened 2025-09-10
Fixed some new errors related to an update in how data is passed to logger and updated packages in response to compromise

jupiter007:
## Description

There was a compromise to some npm packages (https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised). This has been fixed by removing those compromised versions.

- In order to make sure we get the clean packages, I deleted the `package-lock.json` file and re-ran `npm install`.
- Also, due to a change in `@elastic/ecs-pino-format`, we were getting errors because we need to pass the `error` object in differently into `logger.ts`.  So I updated all instances of calls to `logger` to pass the data correctly and updated associated unit tests.

## Type of change

- [X] Chore

### DIFF
--- a/app/api/check-auth/route.ts
+++ b/app/api/check-auth/route.ts
@@ -2,7 +2,6 @@ import { NextResponse } from 'next/server';
 import { getAuthTokenServer } from '@/utils/getAuthTokenServer';
 import { verifyJwtToken } from '@/lib/server/auth';
 import logger from '@/utils/server/logger';
-import { error } from 'console';
 
 const LOGIN = `${process.env.NEXT_PUBLIC_BASE_URL}/login`;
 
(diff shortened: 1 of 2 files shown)


app/api/check-auth/route.ts
```diff
@@ -6,28 +6,28 @@ import logger from '@/utils/server/logger';
...
+    const token = await getAuthTokenServer();
 
+    if (token) {
+      try {
+        const user = await verifyJwtToken(token);
+        if (user) {
+          return NextResponse.json({ authenticated: true });
         } else {
-            return NextResponse.json({ authenticated: false })
+          logger.error('User verification failed');
```
andrewebdev, line 18: All the other calls to `logger.error()` have an object as the first param, and string as the second. This one does not. Is that intentional?

andrewebdev (commented):
Everything looks good, apart from my single comment on one of the `logger.error` calls

PR #805, opened 2025-09-10
Add bash to AWS Dockerfile

briri:
Tried to use session manager to access one of the running containers and could not because bash is not available. Adding it here


jupiter007 (approved):
LGTM

Issue #14, opened 2023-12-18
Build a 'finish' account creation page


briri:
Build out a page that we can direct new users to when they first create an account and we do not get all of their info.

For example if the user authenticates in Cognito with a SAML (or OpenId) based system, and we do not get all of the info we want from that system, they should be redirected to a page asking them for that info.

We want to ensure that we have:
- The correct email address (required)
- The correct affiliation (required)
- The first and last name (required)
- ORCID
- alternate/backup email

Requires **dmsp_rails_prototype** issue [#10](https://github.com/CDLUC3/dmsp_rails_prototype/issues/10)

bofstein:
Should show any information we do have for them but let them edit it, and show the missing fields to fill out

![Image](https://github.com/user-attachments/assets/161ef360-411d-468b-810c-e3f8bb2f850b)

bofstein:
Likely we'll add an SSO redirect to an intermediate page that checks if they have missing information and sends them to an Update form if they're missing information and the dashboard (or wherever we want to send them) if their account is complete.

bofstein:
Do we do this now?  The goal of allowing SSO is that they don't need to create an account.  I would assume we get the required info from the provider and then we wouldn't need this.  We might _want_ them to e.g. provide an ORCID, but making a page like this would be like mandating it, and that would go against the goal of using SSO to avoid account sign up

briri:
Yes, we do this currently and absolutely need this. The only attributes an SSO is "required" to provide are the user's email address and their unique id (aka EPPN)  which within the SSO system. 

Sometimes we get a `Name` field like `Jane Doe` or `Doe, Jane` or `J. Doe`. Other times we get a separate first and last name, other times we get nothing.

Often times, the email address isn't the same as the one the SSO sends us. For example the user might provide `jane.doe@example.com`, get sent to the SSO login, they login and the SSO system sends use `jdoe@example.edu` in the email field. The user entered `jane.doe@example.com` though and either prefers using that or doesn't realize its an alias so we should allow them to use it.

We would show the affiliation/org name in the browser but not allow the user to change this since it is connected to the specific SSO system. 

We would not show them their EPPN as it typically has no value to the user and is only used by us behind the scenes to match what the SSO sends back on subsequent logins to a user in our system. These are sometimes the same as the email address but are typically something like `abcd1234@example.com`. 

We want the user to provide their first and last names, ORCID (maybe?), and confirm the email address that the SSO sent us.

PR #814, opened 2025-09-16
Updates to logging to support OpenSearch

briri:
## Description

Part of [#14](https://github.com/CDLUC3/dmptool-infrastructure/issues/14)

- Added a new `SERVER_LOG_LEVEL` env variable 
- Update `utils/server/logger.ts` to add Pino redaction logic and also a `createLogger` function that allows an optional destination to be passed in to facilitate testing. The default is the normal stdout.
- Updated `utils/server/serverAuthHelper` by adding a `serverFetchAccessToken` to fetch the access token so we can include things like the `jti` in logs
- Added `utils/server/loggerUtils.ts` with a function that will prepare objects for logging. It strips out properties with null/undefined values, and redacts any instances of keys that we don't want in the logs
- Added logging to the `graphqlServerActionHandler.ts` 
- Added some missing `try -> catch` blocks to several nextJS server actions that were missing them
- Updated all server actions to use the new `prepareObjectForLogs` function 
- Added tests for logger and new loggerUtils

## Type of change

## How Has This Been Tested?

All tests are passing. Verified things are working in the UI locally

## Checklist:

### DIFF
--- a/app/[locale]/projects/[projectId]/collaboration/actions/updateProjectCollaboratorAction.ts
+++ b/app/[locale]/projects/[projectId]/collaboration/actions/updateProjectCollaboratorAction.ts
@@ -3,6 +3,8 @@
 import { executeGraphQLMutation } from "@/utils/server/graphqlServerActionHandler";
 import { ActionResponse } from "@/app/types";
 import { UpdateProjectCollaboratorDocument } from "@/generated/graphql";
+import logger from "@/utils/server/logger";
+import { prepareObjectForLogs } from "@/utils/server/loggerUtils";
 
 export async function updateProjectCollaboratorAction({
   projectCollaboratorId,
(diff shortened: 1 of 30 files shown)


app/[locale]/projects/[projectId]/collaboration/actions/updateProjectCollaboratorAction.ts
```diff
@@ -11,12 +13,18 @@ export async function updateProjectCollaboratorAction({
...
-  });
-
+  try {
+    // Execute the mutation using the shared handler
+    return await executeGraphQLMutation({
+      document: UpdateProjectCollaboratorDocument,
+      variables: { projectCollaboratorId, accessLevel },
+      dataPath: "updateProjectCollaborator"
+    });
+  } catch (error) {
```
Comment (jupiter007) [inline app/[locale]/projects/[projectId]/collaboration/actions/updateProjectCollaboratorAction.ts:23]: Don't think we need `try-catch` here either.

app/[locale]/projects/[projectId]/collaboration/invite/actions/addCollaboratorAction.ts
```diff
@@ -23,7 +24,10 @@ export async function addProjectCollaboratorAction({
     });
 
   } catch (error) {
-    logger.error({ error }, `[Add Project Collaborator Error]: ${error}`);
+    logger.error(
```
Comment (jupiter007) [inline app/[locale]/projects/[projectId]/collaboration/invite/actions/addCollaboratorAction.ts:27]: I meant to go back to some of these other actions and remove the `try-catch`. Will do that as part of a separate ticket, so it's ok to just leave these here for now.

jupiter007 (approved):
@briri this looks great. Thanks for doing all this work.

I noticed that we're doing full redaction for the list of REDACTION_KEYS, but I was wondering if we can get away with partial redactions for some of them, like `email` so that they are still sort of searchable if need be. I supposed we can use the `userId` to trace activity, but wasn't sure if a user contacted us about an issue, and all we had was their `email` or `name`, if a partially-redacted value would be easier to troubleshoot.

briri:
ok, updated the `prepareObjectForLogs` function a bit to make sure all emails are masked in the `b****y@ucop.edu` style

PR #556, opened 2025-06-21
Update "required" field notation in living styleguide

fraserclark:
## Description

Added indicator for required fields across all forminput components

Tickets: #503, #552

### Summary of changes

- **Form Components Enhanced**: Added `(required)` text indicators to `FormInput`, `FormTextArea`, `FormSelect`, `RadioGroup`, `CheckboxGroup`, and `TypeAheadWithOther` components

There was an issue with `CheckboxGroup` and `RadioGroup` as there is a isInvalid so I dont think it wasnt using FieldError in react aria properly - so had to do some changes here - let me know if wrong

Styleguide I added form input examples showing both required and non-required states

Test - Added to the test suites including `CheckboxGroup` and `RadioGroup` components

Required text in `<span className="is-required" aria-hidden="true">` - I understand screenreaders might tell the user the form itself is set to required with reactaria so I have put aria-hidden to stop repetition 

### Technical Details

- All form components try to follow consistent React Aria patterns for validation and error handling
- Required indicators are properly styled and accessible (checked the colour too)

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Unit test and manual testing

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I updated the CHANGELOG.md and added documentation if necessary
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have completed manual or automated accessibility testing for my changes
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #844, opened 2025-09-26
Ae 503 required field styling

### LINKED ISSUE #552: Updated "required" field notation in living styleguide
As agreed in meetings, update the "required" indicator on fields in the living styleguide:

- "(Required)" text to appear next to field label, in black (not red)
- Retain the red asterisk

See attached for a reference image. 

Image

andrewebdev:
## Description

This is a new PR based on [#556](https://github.com/CDLUC3/dmsp_frontend_prototype/pull/556)

It adds the new "(required)" field styling to many form input elements, and
allows for setting some files to show the "(required)" text, without being
actually required.

Fixes #552

## Type of change
- [X] Fixed test failing due to new required text

## How Has This Been Tested?

- Manual testing via the living styleguide
- Automated testing

## Checklist:

### DIFF
--- a/components/Form/FormSelect/index.tsx
+++ b/components/Form/FormSelect/index.tsx
@@ -1,4 +1,5 @@
 import React, { forwardRef, Key } from 'react';
+import { useTranslations } from "next-intl";
 import type {
   SelectProps,
   ValidationResult
@@ -52,6 +53,7 @@ export const FormSelect = forwardRef<HTMLButtonElement, MySelectProps<SelectItem
     ...rest
   } = props;
   const showRequired = isRequired || isRequiredVisualOnly;
+  const t = useTranslations('Global.labels');
 
   const handleSelectionChange = (key: Key | null) => {
     if (onChange) {
@@ -74,7 +76,7 @@ export const FormSelect = forwardRef<HTMLButtonElement, MySelectProps<SelectItem
       {(state) => (
         <>
           <Label>
-            {label}{showRequired && <span className="is-required" aria-hidden="true"> (required)</span>}
+            {label}{showRequired && <span className="is-required" aria-hidden="true"> ({t('required')})</span>}
           </Label>
           <Text slot="description" className="help">
             {description}</Text>
--- a/components/Form/FormSelect/index.tsx
+++ b/components/Form/FormSelect/index.tsx
@@ -29,6 +29,8 @@ interface MySelectProps<T extends SelectItem>
   helpMessage?: string;
   description?: string;
   selectClasses?: string;
+  isRequired?: boolean;
+  isRequiredVisualOnly?: boolean;
   onChange?: (value: string) => void;
   items?: T[];
   placeHolder?: string;
@@ -43,11 +45,14 @@ export const FormSelect = forwardRef<HTMLButtonElement, MySelectProps<SelectItem
     helpMessage,
     description,
     selectClasses,
+    isRequired = false,
+    isRequiredVisualOnly = false,
     onChange,
     items,
     placeholder,
     ...rest
   } = props;
+  const showRequired = isRequired || isRequiredVisualOnly;
 
   const handleSelectionChange = (key: Key | null) => {
     if (onChange) {
@@ -63,12 +68,15 @@ export const FormSelect = forwardRef<HTMLButtonElement, MySelectProps<SelectItem
       data-invalid={errorMessage}
       className={`${selectClasses} ${styles.mySelect} react-aria-Select`}
       aria-label={ariaLabel}
+      aria-required={isRequired}
       onSelectionChange={handleSelectionChange}
       placeholder={placeholder}
     >
       {(state) => (
         <>
-          <Label>{label}</Label>
+          <Label>
+            {label}{showRequired && <span className="is-required" aria-hidden="true"> (required)</span>}
+          </Label>
           <Text slot="description" className="help">
             {description}</Text>
           <Button className='react-aria-Button' ref={ref} data-testid="select-button">
@@ -107,13 +115,3 @@ export const FormSelect = forwardRef<HTMLButtonElement, MySelectProps<SelectItem
 });
 
 FormSelect.displayName = 'FormSelect';
-
-
-export function MyItem(props: ListBoxItemProps) {
-  return (
-    <ListBoxItem
-      {...props}
-    />
-  );
-}
-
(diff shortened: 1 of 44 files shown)


components/Form/FormSelect/index.tsx
```diff
@@ -107,13 +114,3 @@ export const FormSelect = forwardRef<HTMLButtonElement, MySelectProps<SelectItem
 });
 
 FormSelect.displayName = 'FormSelect';
-
```
andrewebdev, line 110: I removed this. I checked and it seems the only other place this was used was for tests and in the styleguide. I updated those with appropriate alternatives so that we don't have random code/components around like this.

components/Form/FormTextArea/__tests__/index.spec.tsx
```diff
@@ -13,14 +13,24 @@ describe('FormTextArea', () => {
         label="Name"
         placeholder="Enter your name"
         value="John Doe"
-        onChange={jest.fn()}
```
andrewebdev, line 16: No need to have this here if we are testing it in isolation in another test. Just removed it to keep it clean.

jupiter007:
@andrewebdev it looks good. My main comment was that we need to use translation keys for `(required)`. We also need to check through the `/messages` directory and remove `(required)` from labels there, so we don't get duplicate `(required)(required)` - see some screenshots in my comments.

Also, if you can just take a quick scan over some pages to see if we have fields that are missing the `(required)` text, such as on the `Section name` field of the`Add Section` page.

Lastly, you need to set the `projectName` input field as `isRequired = {true}` on the `/projects/create-project` page, because it currently is set with `aria-required={false}`.

components/Form/TypeAheadWithOther/TypeAheadWithOther.tsx
```diff
@@ -41,9 +43,20 @@ const TypeAheadWithOther = ({
   className,
   suggestions,
   onSearch,
+  isRequired=false,
+  isRequiredVisualOnly=false,
   otherText = "Other",
 }: TypeAheadInputProps) => {
```
andrewebdev, line 50: @jupiter007 I have a weird issue with the translation code below. See the comments in the code. I commented out the non-working code for now, since it was breaking tests, and causing linting errors.

I don't understand why the translation doesn't work here. If you compare it to `FormInput`, I use it exactly the same way, but this one causes an error. (The only difference is that I named this one differently while debugging, since I thought it was conflicting with something else).

components/Form/TypeAheadWithOther/TypeAheadWithOther.tsx
```diff
@@ -197,7 +210,10 @@ const TypeAheadWithOther = ({
         className={(!!error) ? styles.fieldError : ''}
         isInvalid={!!error}
       >
-        <Label>{label}</Label>
+        <Label>
+          {label}
+          {showRequired && <span className="is-required" aria-hidden="true">(required{/*Global('required')*/})</span>}
+        </Label>
```
andrewebdev, line 216: The code above is what throws the error `Global is not a function`.

jupiter007:
> @jupiter007 I've made the changes as requested. It caused a lot of tests to break so I had to fix those also. _However,_ There is a strange bug/issue with the translation inside the `TypeAheadWithOther` component. See my comments above.

Hi @andrewebdev I pulled down your branch. I'm not sure if I have the latest changes because I'm not seeing any failed tests related to `TypeAheadWithOther`. I am, however, seeing two failed unit tests related to `app/[locale]/projects/[projectId]/dmp/create`

andrewebdev:
> > @jupiter007 I've made the changes as requested. It caused a lot of tests to break so I had to fix those also. _However,_ There is a strange bug/issue with the translation inside the `TypeAheadWithOther` component. See my comments above.
> 
> Hi @andrewebdev I pulled down your branch. I'm not sure if I have the latest changes because I'm not seeing any failed tests related to `TypeAheadWithOther`. I am, however, seeing two failed unit tests related to `app/[locale]/projects/[projectId]/dmp/create`

Hi @jupiter007 I've cleared the cache multiple and ran full test suite multiple time just now. With the branch as is, _I don't get any test failures_.

After that I did run `npm generate`, which seems to have changed something, but after I run the tests, again there was _no failures_.

Once I uncomment the 3 lines in `TypeAheadWithOther`, only then do I get around 11 Test failures, all related to that component.

jupiter007:
> > > @jupiter007 I've made the changes as requested. It caused a lot of tests to break so I had to fix those also. _However,_ There is a strange bug/issue with the translation inside the `TypeAheadWithOther` component. See my comments above.
> > 
> > 
> > Hi @andrewebdev I pulled down your branch. I'm not sure if I have the latest changes because I'm not seeing any failed tests related to `TypeAheadWithOther`. I am, however, seeing two failed unit tests related to `app/[locale]/projects/[projectId]/dmp/create`
> 
> Hi @jupiter007 I've cleared the cache multiple and ran full test suite multiple time just now. With the branch as is, _I don't get any test failures_.
> 
> After that I did run `npm generate`, which seems to have changed something, but after I run the tests, again there was _no failures_.
> 
> Once I uncomment the 3 lines in `TypeAheadWithOther`, only then do I get around 11 Test failures, all related to that component.

@andrewebdev I think the shared `useTranslation` mock doesn't work for your changes in `TypeAheadWithOther`.  I added the following the the `TypeAheadWithOther.spec.tsx` file, and the test is passing:
```
// Mock next-intl hooks - override global mock
jest.mock('next-intl', () => ({
  useTranslations: jest.fn(() => {
    const mockT = jest.fn((key: string) => key) as any;
    // Add the rich method that the global mock has
    mockT.rich = jest.fn((key: string, values?: Record<string, any>) => {
      if (values?.p) {
        return values.p(key);
      }
      return key;
    });
    return mockT;
  }),
}));
```

I also cleared all my caches by running `rm -rf .next`, `rm -rf node_modules`, and `npm install` and `npm run generate`, using the latest backend `development` branch, and I still get two unit tests failing for `project/[projectId]/dmp/create` though.

andrewebdev:
@jupiter007 can you paste the errors from the tests that are failing? Try as I might I cannot replicate it. _All_ the tests are passing for me, with zero errors

andrewebdev:
Thanks @jupiter007 ! That does solve the translation issue.

I still cannot produce the other test failures you mentioned above. Can you provide the output you are getting?

PR #850, opened 2025-09-29
Addressed 'fast-redact' vulnerability by updating 'pino' version

jupiter007:
## Description

Dependabot reported a Prototype Pollution vulnerability in `fast-redact` version `3.5.0` and earlier package, where an attacker could potentially inject properties into Object.prototype

Because there is a zero-day bug, there are no patches for this available. However, we updated the `pino` package to a new version that does not use this package.

## Type of change

- [X] Chore

## How Has This Been Tested?
All existing unit tests pass


briri (approved):
LGTM thanks @jupiter007

Issue #710, opened 2025-08-19
Static Pages for Research Outputs table


bofstein:
Build static pages for the flow of adding Research Outputs table as a Question Type

bofstein:
Here is the wireframe I found though I'm not sure if there are any more recent versions: https://www.figma.com/proto/RRtoRLyr5Spf0KiyRfEclt/DMP-Tool-Wireframes?node-id=4554-2010&t=RW5s4TU7JedP9TXZ-0&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=4554%3A2010&show-proto-sidebar=1

PR #896, opened 2025-10-20
Feature/710/js static page for research outputs

jupiter007:
## Description

Created Research Output fields.

- Updated `/template/[templateId]/q/new` page with `researchOutputs` option, even though it hasn't been added yet to `dmptool-types` yet, so that we can simulate the selection of Research Outputs
- For now, I only updated the QuestionAdd page with Research Outputs fields, since a lot might change once we actually implement the Research Outputs backend.
-- Created different components to generate the `License`, `OutputType`, `MetaDataStandards` and `ReposSelector` questions since they are more complicated.
-- Added unit tests

Fixes # ([710](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/710))

## Type of change

## How Has This Been Tested?
This has been manually tested and tested with new unit tests.

## Checklist:

## Important Note:

## To Test
- Create a new template
- At `template/[templateId]` add a new section, and then add a new question
- Select Research Outputs from the question types
- You will get to the "Add new question" form for Research Outputs and be able to interact with the fields there.

## Screen recording of new functionality
This is for the QuestionAdd page. A user selects Research Output question type to edit the new fields.

This is a static page, and the backend work has not been done yet, so no data is actually being saved.

https://github.com/user-attachments/assets/7d3e6c84-52db-4d27-ab8d-8de540c27726

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,6 +81,7 @@
       "version": "3.13.8",
       "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.13.8.tgz",
       "integrity": "sha512-YM9lQpm0VfVco4DSyKooHS/fDTiKQcCHfxr7i3iL6a0kP/jNO5+4NFK6vtRDxaYisd5BrwOZHLJpPBnvRVpKPg==",
+      "peer": true,
       "dependencies": {
         "@graphql-typed-document-node/core": "^3.1.1",
         "@wry/caches": "^1.0.0",
@@ -249,6 +250,7 @@
       "version": "7.28.4",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.4.tgz",
       "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
@@ -1222,6 +1224,7 @@
           "url": "https://opencollective.com/csstools"
         }
       ],
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -1244,6 +1247,7 @@
           "url": "https://opencollective.com/csstools"
         }
       ],
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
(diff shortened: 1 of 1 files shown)


app/types/index.ts
```diff
@@ -2,6 +2,9 @@ import { ReactNode } from "react";
 import { PlanSectionProgress, TemplateVisibility, PlanFeedback, ProjectFundingStatus } from "@/generated/graphql";
 import { AffiliationSearchQuestionType, AnyQuestionType } from '@dmptool/types';
 
+// Re-export types from questionAdd module
```
jupiter007: Going to try and organize the `app/types/index.ts` going forward, so I created a separate `types` file for just `questionAdd` types for this ticket.

components/QuestionAdd/LicenseField.tsx
```diff
@@ -0,0 +1,134 @@
+import { Button, ListBoxItem } from "react-aria-components";
+import { useTranslations } from 'next-intl';
```
jupiter007: New static page for the Research outputs License field

components/QuestionAdd/Selector.module.scss
```diff
@@ -0,0 +1,402 @@
+@use '@/styles/responsive' as r;
+
```
jupiter007: Styles for the metadata standards and repo selector fields

components/AdminSectionEditContainer/__tests__/index.spec.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { act, render, screen } from "@testing-library/react";
+import { act, render, screen, fireEvent, waitFor } from "@testing-library/react";
```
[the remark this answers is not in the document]
briri: 😎 thanks for adding this test

jupiter007:
> This is looking really good @jupiter007
> 
> One thing I think we should do though and maybe we should wait for @bofstein to make a final decision (up to you if you want to do it in this PR or wait until we wire the page to the backend). When the user expands a section and makes a change it should automatically check the box for that section (e.g. if I expand repos and add a preferred repository the checkbox that indicates that we want to include it in the question should get auto-checked.

@briri would there ever be a case where a user modifies a field but doesn't want it included or active yet in the Research Output? If we are always checking the checkbox for a field when a user makes updates, then maybe we don't need the checkboxes at all, because it seems redundant. I'll wait to run this by Becky when she is back.

Thanks for your review Brian :)

PR #921, opened 2025-10-30
[Chore] - Updated version of cypress to 14.5.4 and fixed broken cypress tests a…

jupiter007:
## Description

- Updated version of cypress to 14.5.4
- Fixed broken cypress tests
- Moved loading of tinymce out of <head> tags in `layout.tsx`, to TinyMCEEditor, so that it will be lazy-loaded only when used.

### DIFF
--- a/utils/loadTinyMCE.ts
+++ b/utils/loadTinyMCE.ts
@@ -6,9 +6,16 @@ export function loadTinymceScript(): Promise<void> {
   if (window.tinymce) return Promise.resolve();
   if (tinymceLoadPromise) return tinymceLoadPromise;
 
-  tinymceLoadPromise = new Promise((resolve, reject) => {
+  tinymceLoadPromise = new Promise((resolve, reject) => { //singleton pattern - subsequent calls in same page return the same promise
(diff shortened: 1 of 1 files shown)


app/[locale]/layout.tsx
```diff
@@ -56,13 +55,7 @@ export default async function LocaleLayout({
 
   return (
     <html lang={locale} className={font_sans_serif.variable}>
-      <head>
```
Comment (jupiter007) [inline app/[locale]/layout.tsx]: Removed this from <head> because even with the `beforeInteractive` which loads it early in the page lifecycle, it was still loading on every page, and most pages don't use TinyMCEEditor.

components/Header/index.tsx
```diff
@@ -39,6 +39,7 @@ function Header() {
   }, [isAuthenticated]);
 
   const handleLogout = async () => {
+    setShowMobileMenu(false);
```
jupiter007: I know this Header is just a temporary filler, but noticed that the mobile menu doesn't close when a user clicks on the "Logout" button in the mobile navigation. This was causing the "logout" cypress test to fail.

PR #1156, opened 2026-02-05
[Chore] - Updated renovate.json with schedule and minimumReleaseAge, and added Trivy to build pipeline

jupiter007:
## Description

Updated `renovate` config to include a schedule and minimumReleaseAge and rangeStrategy=pin.

- Updated `renovate.json` to pick up on updates in `docker-compose.yaml` and added a schedule and `minimumReleaseAge`.


jupiter007:
Removed `trivy`-related code in buildspec.yaml

PR #1208, opened 2026-03-18
Fixed issue where output type answer was being saved using id rather than value

jupiter007:
## Description
One more small bug I ran across while working with the Research Output data. 

- Fixed issue where selected output type on Research Output answer was being saved using id rather than the value, so that the `answer` json will not contain something like `my-output-type`, but rather `my output type`, and that is what will display in the answer.

Fixes # ([141](https://github.com/CDLUC3/dmptool-doc/issues/141))

## Type of change

## How Has This Been Tested?
Manually tested

## Checklist:

## Important Note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 - Updated `/login` page step=email to include a "Back" button so user can go back to Step 1 [#997]
 
 ## Fixed
+- Fixed issue where selected output type on Research Output answer was being saved using id rather than the value
 - Fixed some issues in `Research Output` question answer form, where data flags checkboxes were not displaying and output types was missing the custom output
 - Fixed issue with missing `as builder` in `Dockerfile.prod`
 - Fixed bug where user's org was not displaying as a filter option, and Best practice was displaying instead [#62]
--- a/components/Form/ResearchOutputAnswerComponent/SingleResearchOutputComponent/index.tsx
+++ b/components/Form/ResearchOutputAnswerComponent/SingleResearchOutputComponent/index.tsx
@@ -518,11 +518,18 @@ const SingleResearchOutputComponent = ({
                     isRequired={col.required}
                     name={name}
                     options={tooltipSelectItems}
-                    defaultSelectedKey={String(value)}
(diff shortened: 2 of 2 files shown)


briri (changes requested):
Is it possible to just add some code to the narrative generator to convert `some-value` id to `Some value` label when generating the narrative output?

We want to retain the id here. When the answer is displayed in the UI doesn't it need to match up to the list item using the id? Also, this is used in the common standard so we want to retain the id.

Another option I suppose would be to update the `@dmptool/types` to allow both the id and label to be provided.

jupiter007:
> Is it possible to just add some code to the narrative generator to convert `some-value` id to `Some value` label when generating the narrative output?
> 
> We want to retain the id here. When the answer is displayed in the UI doesn't it need to match up to the list item using the id? Also, this is used in the common standard so we want to retain the id.
> 
> Another option I suppose would be to update the `@dmptool/types` to allow both the id and label to be provided.

Good point @briri. I'm not sure about converting "some-value" to "Some value", as there could be a type with a hyphen, especially if users can add a custom one.

But I do like the idea of updating the `@dmptool/types` to allow both id and label to be provided. I will create a ticket for that and close this ticket.

jupiter007:
> > Is it possible to just add some code to the narrative generator to convert `some-value` id to `Some value` label when generating the narrative output?
> > We want to retain the id here. When the answer is displayed in the UI doesn't it need to match up to the list item using the id? Also, this is used in the common standard so we want to retain the id.
> > Another option I suppose would be to update the `@dmptool/types` to allow both the id and label to be provided.
> 
> Good point @briri. I'm not sure about converting "some-value" to "Some value", as there could be a type with a hyphen, especially if users can add a custom one.
> 
> But I do like the idea of updating the `@dmptool/types` to allow both id and label to be provided. I will create a ticket for that and close this ticket.

Ok, I created this ticket: https://github.com/CDLUC3/dmptool-doc/issues/142 for `@dmptool/types`

jupiter007:
Closing, as we will take a different approach to solving the issue with Research Output `output types` displaying with hyphens

PR #1268, opened 2026-05-08
[chore] - Addressed vulnerabilities with next-intl, icu-minify and postcss

jupiter007:
## Description

- Updated version of `next-intl` to `v4.9.2` and `icu-minify` to `v4.11.1`, and added `override` of `postcss` to `v8.5.10` to address security vulnerabilities.


briri (approved):
LGTM

PR #1302, opened 2026-06-10
Added Github Action workflows to automate versioning

jupiter007:
## Description

I was going to push up directly to `main` branch, but it has some `type` errors, so I thought I'd just create this PR and then plan on merging to `stage` and then merging `stage` to `main`. I want to test this workflow on `dmptool-ui` before applying it to all the repos.

Essentially, the `workflows` will check for `version:major`, `version:minor`, or `version:patch` labels on the PR. If they are not present, it will default to a `patch` version.

The workflow will bump `package.json`, prepend the new version at the top of `CHANGELOG.md`, commit files, and push a git tag to `stage`.

- Added Github Action workflows `versioning.yml` and `validation-version-label.yml` to automate versioning when merging from `development` into `stage`.
- Updated `README.md`

Fixes # ([271](https://github.com/CDLUC3/dmptool-doc/issues/271))

## Type of change
- [X] Chore

### DIFF
--- a/.github/workflows/versioning.yml
+++ b/.github/workflows/versioning.yml
@@ -4,7 +4,7 @@ on:
     types:
       - closed
     branches:
-      - stage
+      - main
 
 permissions:
   contents: write
@@ -66,6 +66,10 @@ jobs:
             echo "### Updated"
             echo ""
             echo "### Fixed"
+            echo ""
+            echo "### Removed"
+            echo ""
+            echo "### Chore"
             echo ""
             cat CHANGELOG.md
           } > CHANGELOG.tmp
--- a/.github/workflows/versioning.yml
+++ b/.github/workflows/versioning.yml
@@ -1,0 +1,87 @@
+name: Automated Versioning
+on:
+  pull_request:
+    types:
+      - closed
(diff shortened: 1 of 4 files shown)


.github/workflows/versioning.yml
```diff
@@ -0,0 +1,87 @@
...
+          VERSION="${{ steps.version.outputs.version }}"
+
+          {
+            echo "## v${VERSION}"
+            echo ""
+            echo "### Added"
+            echo ""
+            echo "### Updated"
+            echo ""
+            echo "### Fixed"
```
briri, line 68: Nice!
Do we want to add `### Removed` as well?

briri (approved):
Looks good @jupiter007

.github/workflows/versioning.yml
jupiter007: @briri that makes sense. I also added ### Chore. 

I also updated the workflows to be just for merge to `main`. I'm thinking that makes the most sense so the versioning will remain accurate. 

I'm still not convinced we need release candidate versioning, but we can see how it goes.

PR #1314, opened 2026-06-16
Fixed bug in logo upload

briri:
Was testing in dev and realized that there was a bug when trying to add a logo to an org that does not already have one defined.

The `logoName` variable was overloaded. Its defined using `useState` but we were also defining it locally within the `updateAffiliation` function which was causing confusion. The bug was resulting in the `logoName` never getting updated when the existing record had no prior `logoName`.


jupiter007 (approved):
LGTM

Issue #25, opened 2024-04-04
Complex question type: Repository selector 


briri:
We need to define what we want the repository selector to look like and its overall functionality.

The current DMPTool's repo selector looks like this:
<img width="1056" alt="Screenshot 2024-04-04 at 10 18 05 AM" src="https://github.com/CDLUC3/dmsp_frontend_prototype/assets/1204467/efefb475-de41-4847-af94-f21dd463458b">

It offers some basic filters and a search box. The template admin is able to customize the initial list that the user is presented with (preferred repos). The user selects a repo by clicking the 'Select' button. The smaller blue pills contain keywords related to that repo. Clicking on one of those pills presents the user with all of the repos that use that keyword.

Once the user selects an item it appears on the underlying research output page:
<img width="1033" alt="Screenshot 2024-04-04 at 10 17 34 AM" src="https://github.com/CDLUC3/dmsp_frontend_prototype/assets/1204467/55487e73-7587-4077-bd77-8f2a26ea7f49">

The user can 'Remove' the repo from this page or click the button to add more repos.

In the new system we want:
- Retain the functionality that allows the template author to define preferred (or required) repositories. For example some institutions may require that a copy of the research output be placed into their institutional repository. The user would still be able to submit to other repos in that scenario.
- Provide more robust filter/facet controls along with the search box
- Consider other information provided elsewhere in the plan creation process. For example if they specify that their project is in the `Biology` research domain, we could pre-filter the repository list to show repositories that specialize in that area. Or (a more advanced thing) if the funder is NIH and we know they require the outputs to have a DOI and certain access methods, we can tap into the repository's extended metadata to prefer or somehow highlight a repository as being compliant with that funder's needs.
- MIT has a good example we can learn (borrow) from:
  - A page discussing some of the considerations/guidance for selecting: https://libraries.mit.edu/data-management/share/find-repository/
  - Here is their finder tool: https://data-repository-finder.ll.mit.edu
- [Re3data](https://www.re3data.org/search) itself (where we get the repository list and metadata from) is also a good source for inspiration: 
   
See the corresponding [AWS infrastructure ticket](https://github.com/CDLUC3/dmsp_aws_prototype/issues/86)

jupiter007:
Thanks @briri. This almost seems like a UI/UX ticket, and subsequent work for frontend can be blocked by this one. Should we add a label for UI/UX work?

briri:
Yeah I think that's a good idea

[...]

PR #1344, opened 2026-07-10
Bug/common standard

briri:
## Description

Fixes [#284](https://github.com/CDLUC3/dmptool-doc/issues/284)

**See corresponding PRs in narrative generator, apollo server, and the `@dmptool/types` repos**

- Update both Question and Answer components for `researchOutputTable` types to use the `commonStandardId`

Also fixes [#25](https://github.com/CDLUC3/dmptool-doc/issues/25)
- Add "Anticipated Release Date" and "Anticipated File Size" to the add question page so the admin can enable/disable the field and provide their own help text

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Verified that I was able to create/update/delete research output type questions, publish the template, create/update plans based on the new template

## Checklist:


components/Form/ResearchOutputAnswerComponent/SingleResearchOutputComponent/index.tsx
```diff
@@ -847,88 +826,99 @@ const SingleResearchOutputComponent = ({
...
+                >
+                  {(item) => <ListBoxItem key={item.id}>{item.name}</ListBoxItem>}
+                </FormSelect>
+              </div>
+            );
+          }
         }
       })}
 
-      {/* Always include Anticipated Release Date - not configurable */}
```
jupiter007, line 866: Is "Anticipated Release Date" and "Anticipated File Size" only displayed when it is selected in the original Template? I didn't see an option to select those when creating a template.

components/Form/ResearchOutputAnswerComponent/SingleResearchOutputComponent/index.tsx
```diff
@@ -214,15 +214,6 @@ const SingleResearchOutputComponent = ({
     setErrors({});
```
jupiter007, line 250: L135: For "validateField", do we now need to add radioButtons, numberWithDate, and numberWithContext to cover Initial Access Level, Anticipated Byte Size and Anticipated Release Date?

jupiter007 (approved):
Looks good Brian. I just had a couple of questions.

jupiter007:
@briri something else I did notice is that I was not able to remove or override the "0" in the "Anticipated File Size" field. Also, the default text in the Anticipated File Size field was wrapping. The width of that field should be extended to prevent the wrapping.

components/Form/ResearchOutputAnswerComponent/SingleResearchOutputComponent/index.tsx
briri: We never had them there. I don't recall why, so I left that as-is. They just always appear to the user when answering a new row on the question.

components/Form/ResearchOutputAnswerComponent/SingleResearchOutputComponent/index.tsx
briri: yeah, we probably should. I'll add that logic in

briri:
Thanks for review @jupiter007.

I added some field validation for those other fields. I'm not sure how to appropriately fix the sizing of the select box though. It only looks like that when you add a new row (if you edit an existing its ok) and from what i can see its using the same component. Let me know if you have any advice

jupiter007:
> Thanks for review @jupiter007.
> 
> I added some field validation for those other fields. I'm not sure how to appropriately fix the sizing of the select box though. It only looks like that when you add a new row (if you edit an existing its ok) and from what i can see its using the same component. Let me know if you have any advice

I will take a look tomorrow and give you the adjustment

jupiter007:
Brian, looks like you have to set `flex:1` for the unit dropdown so it can extend the full width. You could set a max-width on it to keep it from getting too wide though:
```
.fileSizeSelect {
  flex: 1;
  max-width: 250px;
}
```

briri:
Thanks @jupiter007 I updated that style and it fixed the issue.

I also made some modifications to a few of the files to get the "Anticipated Release Date" and "Anticipated File Size" into the template QuestionAdd so the admin can now enable/disable those fields and provide their own help text for each. This is for [#25](https://github.com/CDLUC3/dmptool-doc/issues/25)

jupiter007:
@briri I'm still seeing this one issue where I see a "0" in the `Anticipated File Size` field, and I can't override it with any entry: 
image

jupiter007:
> @briri I'm still seeing this one issue where I see a "0" in the `Anticipated File Size` field, and I can't override it with any entry: image

If you look on Dev, we were able to enter any number before.

jupiter007:
> Thanks @jupiter007 I updated that style and it fixed the issue.
> 
> I also made some modifications to a few of the files to get the "Anticipated Release Date" and "Anticipated File Size" into the template QuestionAdd so the admin can now enable/disable those fields and provide their own help text for each. This is for [#25](https://github.com/CDLUC3/dmptool-doc/issues/25)

Looks great @briri. I just had the one concern about not being able to enter a value for the Anticipated File Size field.

briri:
I'm not sure what you mean @jupiter007. I'm able to enter numbers into that field. I'm not sure why its showing a zero in the field by default I didn't really change much on that `FormInput` component beyond the handler event

jupiter007:
> I'm not sure what you mean @jupiter007. I'm able to enter numbers into that field. I'm not sure why its showing a zero in the field by default I didn't really change much on that `FormInput` component beyond the handler event

Ok, if it's working for you then it's not an issue. I'll have to check back later when I remember. I know that Fraser made some changes recently to the Number component. It could be related. Thanks.

PR #1346, opened 2026-07-13
Fix typeahead clearing affiliation on focus and improve both  autocomplete accessibility

### LINKED ISSUE #80: Decide and integrate on a Icon set or Library (FontAwesome or Google Material)

fraserclark:
## Description

Fixes #80
https://github.com/CDLUC3/dmptool-doc/issues/83

On profile edit, clicking into the institution field wiped whatever was already there. You had to type the whole name again just to fix a typo or pick a different subschool. The cause was in the shared typeahead: `handleInputClick` deliberately cleared the value and form data on every click.

Fixed that in `TypeAheadWithOther` (and the older `TypeAheadInput` copy for consistency). Default now keeps the text and puts the caret at the end. If a screen ever needs the old clear-on-focus behaviour, there's a `clearOnFocus` prop.

Since I was already in the component:
- `useId()` for listbox/option IDs (we had hardcoded `id="results"` which breaks when two typeaheads are on one page when I was testing states)
- Suggestion highlight colours swapped to the same `--highlight-*` tokens as FormSelect — the old grey on dark text was tough to read - maybe not accessible.

Wired `isLoading={isSearching}` on the pages that use the hook (profile, signup, add funding, member search, admin user manage, affiliation question component).

https://github.com/user-attachments/assets/dbfe4b96-2285-4e01-9a42-467523b120d8

## Type of change

- [ ] Chore

## How Has This Been Tested?
 Automated jest/tests ran and manual tested on various pages where we use typeahead

## Checklist:

## Important Note:

Merged development before opening this.


components/Form/TypeAheadWithOther/useAffiliationSearch.ts
```diff
@@ -7,36 +7,74 @@ import { AffiliationsDocument } from '@/generated/graphql';
 // Utils and other
 import { debounce } from '@/hooks/debounce';
 import { SuggestionInterface } from '@/app/types';
+import { handleApolloError } from '@/utils/apolloErrorHandler';
+
+const SEARCH_ERROR_MESSAGE = 'Could not load institution suggestions. Please try again.';
```
jupiter007: L12: This should be using a translation key

jupiter007 (commented):
Looks good to me. Please just add a translation key for the `SEARCH_ERROR_MESSAGE`

jupiter007 (approved):

PR #1350, opened 2026-07-16
Fixed issue with local ref to @dmptool/types 

briri:
Looks like I did the same think @jupiter007 

this fixes the local reference to `@dmptool/types` and also adds an override for a new security vulnerability


jupiter007 (approved):
LGTM

